PR #6, opened 2018-11-29
Release 0.5

accuminium:


src/axi_test.sv
```diff
@@ -15,6 +15,11 @@
 
 import axi_pkg::*;
 
+interface AXI_CLK (
```
zarubaf: Are we gaining anything by using this one-bit interface? If not I would suggest replacing it with a `logic`.

fabianschuiki: We think it's the only way to pass a signal into a class. If we would use a logic in the constructor of the class the value itself would be stored, rather than a pointer to the signal which we could use to wait for a posedge. Unless we understood the standard wrong, in which case we should do what you suggest!

fabianschuiki:
Minor bumps in v0.y.z are considered [breaking changes](
https://semver.org/#spec-item-4) in semantic versioning. The library used by Bender to do the semver business should also treat this as a breaking change.

src/axi_test.sv
zarubaf: You are most probably right about this. I still would suggest to copy the entire interface and not make a dummy clock-only interface, this feels really hacky. Imho, this solution doesn't scale really well (e.g.: switching to a more elaborate verification infrastructure with clocking blocks or multiple clock domains). So we can have an AXI interface which is meant for testing and one which we can be used for implementation. Ideally we would get rid of the RTL AXI interface piece by piece and replace it by structs (for example https://github.com/pulp-platform/ariane/blob/master/include/ariane_axi_pkg.sv).

zarubaf: You most likely would also include the reset signal in the verification interface.

zarubaf:
Awesome, good to know.

zarubaf:
Sorry to be picky about that, from the link you send: `MINOR version when you add functionality in a backwards-compatible manner, and` Removing ports from a module/interface will break and not be backwards compatible (e.g.: considering the Kosmodrom TL for example). Never mind if bender doesn't care, I can live with only a minor bump :)

accuminium:
Sure, this is a breaking change, but in SemVer 2.0.0, item 4 is

> 4. Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

and item 8 explicitly mentions `X > 0`:

> 8. Major version X (X.y.z | X > 0) MUST be incremented if any backwards incompatible changes are introduced to the public API. (...)

That's why I suggest to tag this v0.5. If you prefer to tag it v1.0 and thus start a more stable release series, that's of course also fine for me.

src/axi_test.sv
accuminium: On structs in AXI, I clearly see an advantage of having structs like in `ariane_axi` because you can easily form variations of the full AXI interface by composition. However, there is no way to parametrize structs, right? So how would you represent channels with different ID, data, and address widths?

On the verification interface, how would you connect an AXI verification interface (with clock and possibly reset) to a regular interface without wrapping and unwrapping?

zarubaf: Imho you don't even want to provide the ability to parameterise them (at least not the way it is done right now). The interfaces lead you to the (often wrong) conclusion, that you can arbitrarily parameterise your interface.
- If you really have a module which can operate on all kinds of AXI Buses you could pass the type of the `req`, `resp` as a parameter (as done in the stream arbiter, etc...), a crossbar, etc.
- Regarding ID width, I really don't see the use-case of arbitrary ID width. The AXI standard **recommends** a master ID width of 4 which provides, more often than not, enough out-of-order streams for masters.
- We could settle on an agreed slave ID. I'd say that there is only a limited amount of slaves which can be attached to a xbar (e.g..: 64 - 128?). Constant propagation would prune the rest. 
- Address width is an easy conversion.
- Data width actually needs size converters. So you will need a up/down converter anyway. 
- The AXI package could provide commonly used AXI incarnations.

The test-bench usually connects with interfaces on the top-level ports. If you are lazy you could write a macro for that. 

On the other hand you gain:
- Clear error reporting (no empty interfaces segfaults).
- Clear direction of buses, even the tool can't figure it out sometimes.
- No top-level `interface <-> logic` wrappers. 
- The ability to use `assign` between structs and tying-off.
- You would actually be compliant with the (maybe/hopefully) coming style guide.
- Nice signal grouping in Modelsim ;-)

I understand that this is absolutely out-of-scope of this PR, but maybe we can avoid introducing yet another interface without having a clear advantage.

zarubaf: We could move the assignment between interface and structs to a macro in the `axi_pkg`. Otherwise I would say it is up to the TB implementer to take care of this.

Note that with the current changes we will have breaking changes again. I'd suggest to make a hard-copy of the `AXI_BUS` with and without clock. The one with the clock can be used during verification and on TB level. The one without the clock can be used for RTL implementation and can slowly be phased out when appropriate. I would like to avoid that we need to re-structure all TB using this interface again (e.g.: separate clock vs. clock in the interface). Then I am ok with this change and I can understand the benefit.

zarubaf: Also the assignment to connect `AXI_BUS` to `AXI_BUS_DV (clk_i)`.

accuminium: I see the benefits of your suggestion, but I think we will need
- some practical examples of actual use cases to guide the implementation
- a set of structs that implement the most relevant configurations in this package
- some preprocessor support (e.g., macros) to reduce boilerplate
- guidelines / dos&donts for creating 'struct ports' and connecting modules using them
- a strategy / guide / scripts to transition existing modules from interfaces to struct ports

to definitely make the switch.

Maybe we can start drafting this for v1.0?

accuminium:
So in conclusion: In the long run we want to switch to structs (e.g., for v1.0), and in the short run we accept this MR as is?  Or should something specific be changed?

zarubaf:
Those two things are really orthogonal. What I don't accept in this pull request is the separate clk interface. As said, my suggestion would be to duplicate the interface: one with a clock and one without a clock. Then you can use the one without clock in your design and the one with the clock in your TB infrastructure. The changes here are quite hacky.

fabianschuiki:
Sounds good!On 5 Dec 2018 18:07, Andreas Kurth <notifications@github.com> wrote:So in conclusion: In the long run we want to switch to structs (e.g., for v1.0), and in the short run we accept this MR as is?  Or should something specific be changed?

accuminium:
Thanks. I will merge this after v1.10 of `common_cells` has been released.

PR #9, opened 2019-05-15
Remove interface name aliasing to keep VCS happy

jrrk:
At the moment in an alias for Slave and out an alias for Master. This causes VCS to complain and refuse to compile when the names are mixed. This PR remove the in and out aliases. Linked to

PR #11, opened 2019-05-24
Remove interface name aliasing to keep VCS happy

accuminium:
Continuation of #9.


accuminium:
Synthesis and simulation regressions pass.

accuminium:
As this is a breaking interface change, it will go into v0.7.

accuminium:
Thanks!

PR #15, opened 2019-06-03
WIP: Various improvements for v0.8

accuminium:
### Added
- `axi_id_remap` now raises an error when it receives any atomic operation, which it currently does
  not support.
- Add macros for assigning to and from AXI channel structs and an AXI interface.

### Removed
- `axi_pkg` previously defined fixed widths for ID, user, address, and data signal and derived
  structs for the five AXI channels from them.  However, these widths are not equal for all
  instantiations and usages of this package.  Therefore, they have been removed.

### TODO
- [ ] Add overview table of modules to README (@accuminium).
- [ ] Add struct templates (@zarubaf).
- [ ] Add connector modules (@zarubaf).


accuminium:
The transition to structs has been implemented in #28, and `axi_id_remap` has been removed (will soon be replaced by a correct implementation).

Issue #21, opened 2019-10-26
Test driver compliance with AXI spec


chiraag:
The current test-driver in `axi_test.sv` waits for `axi.aw_ready` to be high in the `send_aw` task before trying the `send_w` task. I think this test-driver is not compliant with the AXI spec. AXI4 write response dependency sub-section in section A3.3 of the spec says that 

> the master must not wait for the slave to assert AWREADY or WREADY before asserting AWVALID or WVALID

> the slave can wait for AWVALID or WVALID, or both, before asserting AWREADY

Thus the test-bench can potentially fail an otherwise correct crossbar implementation. I might try to edit the test-driver to fix this and was wondering if you have any thoughts on a simple fix.

accuminium:
Hi Chiraag,

Thank you for your report.

Can you please link the behavior you are describing to code?  I understand that [the `send_aw` task of `axi_driver` waits for `aw_ready`](https://github.com/pulp-platform/axi/blob/d252bdf210cdf5e57b0cda3346739aedc7ac5f36/src/axi_test.sv#L375) before returning and therefore that waiting for `send_aw` to finish before calling `send_w` would violate the AXI specification.  However, `axi_driver` does not invoke these tasks itself but delegates that to users.

So the code in `axi_test.sv` is correct but it might be used incorrectly, right? If so, where?

fabianschuiki:
Indeed the "correct use" is currently left to the user. The `axi_test` driver only provides a simple interface to perform per-channel transactions. Maybe it would be a good idea to provide a `write` and `read` task which operate on an unsized queue `data_t data[$]` and send an entire sequence of data. In that task we could then also implement the dependency graph of AXI beats. What do you think, @accuminium -- is this a use case?

accuminium:
Yes, and there are developments towards higher-level AXI test drivers, e.g., [`rand_axi_master`](https://github.com/pulp-platform/axi/blob/190e80547d354857d4a7949ab472f647c5219074/src/axi_test.sv#L580). I think the driver you describe would fit nicely between the current, low-level `axi_driver` and the very high-level `rand_axi_master`. The instantiation chain could then be `rand_axi_master -> axi_master -> axi_driver` (`->` = instantiates), and users could instantiate one of these modules according to their required level of control. What do you think?

fabianschuiki:
This sounds excellent!

accuminium:
Perfect. Contributions welcome!

chiraag:
[...]

PR #22, opened 2019-10-28
tb_axi_lite_xbar: Fix AW->W dependency and increase handshake randomness

accuminium:
Fixes the problem reported in #21.


accuminium:
Our Lite crossbar still passes this TB.

chiraag:
This works for my lite cross-bar too

PR #25, opened 2019-12-03
Fix axi to axi lite

zarubaf:
It seems that I've fixed underflow conditions here. I guess that was in conjuction with eh billywig tape-out.


accuminium:
Thank you for these fixes!

I have rebased on `master` and added an entry to the Changelog.

PR #26, opened 2019-12-03
Add AXI CDCs

zarubaf:
This adds AXI CDCs which are based on the safer FIFO Gray design from our `common_cells` library.

Todo:
- [x] Rebase on xbar implementation
- [x] Run testbench
- [x] Add changelog entry
- [x] Add interface wrapper


zarubaf:
That is perfect. Thanks a lot!

PR #28, opened 2019-12-19
Transition from interfaces to struct ports

accuminium:


accuminium:
Any objections on releasing v0.8 with struct ports?

fabianschuiki:
> Any objections on releasing v0.8 with struct ports?

Not at all, fine by me!

src/axi_multicut.sv
```diff
@@ -10,65 +10,237 @@
...
+    .slv_req_i  ( slv_req  ),
+    .slv_resp_o ( slv_resp ),
+    .mst_req_o  ( mst_req  ),
+    .mst_resp_i ( mst_resp )
   );
 
+  // Check the invariants.
+  // pragma translate_off
+  `ifndef VERILATOR
+  `ifndef VCS
```
zarubaf: Why are we excluding VCS?

src/axi_multicut.sv
```diff
@@ -10,65 +10,237 @@
...
+  `ifndef SYNTHESIS
+  initial begin
+    assert ( ADDR_WIDTH > 0 ) else $fatal(1, "Wrong addr width parameter");
```
zarubaf: I am in favor of removing the spaces around the parentheses.
```suggestion
    assert (ADDR_WIDTH > 0) else $fatal(1, "Wrong addr width parameter");
```

src/axi_multicut.sv
```diff
@@ -10,65 +10,237 @@
...
+    assert ( DATA_WIDTH > 0 ) else $fatal(1, "Wrong data width parameter");
```
zarubaf: Same for below.

src/axi_multicut.sv
```diff
@@ -10,65 +10,237 @@
...
+    .slv_resp_o ( slv_resp ),
+    .mst_req_o  ( mst_req  ),
+    .mst_resp_i ( mst_resp )
   );
 
+  // Check the invariants.
+  // pragma translate_off
+  `ifndef VERILATOR
+  `ifndef VCS
+  `ifndef SYNTHESIS
```
zarubaf: That should be already covered with `pragma translate_off`.

src/axi_multicut.sv
```diff
@@ -10,65 +10,237 @@
...
+
+    // connect master to the highest index
+    assign mst_req_o        = cut_req[NoCuts];
+    assign cut_resp[NoCuts] = mst_resp_i;
+  end
+
+  // Check the invariants
+  // pragma translate_off
+  `ifndef VERILATOR
+  `ifndef VCS
```
zarubaf: Same as above.

src/axi_multicut.sv
```diff
@@ -10,65 +10,237 @@
...
+    .slv_req_i  ( slv_req  ),
+    .slv_resp_o ( slv_resp ),
+    .mst_req_o  ( mst_req  ),
+    .mst_resp_i ( mst_resp )
+  );
+
+  // Check the invariants.
+  // pragma translate_off
+  `ifndef VERILATOR
+  `ifndef VCS
```
zarubaf: Same as above.

src/axi_multicut.sv
accuminium: Good question. It is only being excluded in `axi_cut` and `axi_multicut`, so I am removing it. If there really are issues, we have to exclude VCS for all modules.

accuminium:
Great, thanks!

PR #35, opened 2020-01-13
Re-work AXI to AXI lite

WRoenninger:


accuminium:
What is the status of this? Can it be rebased on the current master with the AXI-Lite verification infrastructure and then reviewed?

WRoenninger:
Yes should be possible to rebase it on the current master

src/axi_to_axi_lite.sv
```diff
@@ -9,234 +9,317 @@
...
-    assert(NumPendingRd > 0) else $fatal(1, "Number of pending reads must be greater than 0!");
-    assert(NumPendingWr > 0) else $fatal(1, "Number of pending writes must be greater than 0!");
-  end
-`endif
-// pragma translate_on
 
+  // Assertions
+  // pragma translate_off
+  `ifndef VERILATOR
+  `ifndef VCS
```
zarubaf: Why is the `VCS` here?

src/axi_to_axi_lite.sv
```diff
@@ -9,234 +9,317 @@
...
-`ifndef VERILATOR
-  initial begin: p_assertions
-    assert (AXI_ADDR_WIDTH >= 1) else $fatal(1, "AXI ADDR width must be at least 1!");
-    assert (AXI_DATA_WIDTH >= 1) else $fatal(1, "AXI DATA width must be at least 1!");
-    assert (AXI_ID_WIDTH   >= 1) else $fatal(1, "AXI ID width must be at least 1!");
-    assert (AXI_USER_WIDTH >= 1) else $fatal(1, "AXI USER width must be at least 1!");
+  // Assertions, check params
+  // pragma translate_off
+  `ifndef VERILATOR
+  `ifndef VCS
```
zarubaf: Same here, do we need the `VCS` guard?

WRoenninger:
Added suggestions to PR

test/tb_axi_to_axi_lite.sv
```diff
@@ -9,7 +9,6 @@
 // specific language governing permissions and limitations under the License.
 //
 // Fabian Schuiki <fschuiki@iis.ee.ethz.ch>
-// Andreas Kurth  <akurth@iis.ee.ethz.ch>
```
accuminium: Why are you removing an author? Feel free to add yourself, though.

src/axi_to_axi_lite.sv
```diff
@@ -9,234 +9,318 @@
...
-  input  logic        rst_ni,
-  input  logic        testmode_i,
-  // Slave AXI port
-  input  req_t        slv_req_i,
-  output resp_t       slv_resp_o,
-  // Master AXI-Lite port
-  output req_lite_t   mst_req_o,
-  input  resp_lite_t  mst_resp_i
+	input  logic       clk_i,    // Clock
+	input  logic       rst_ni,   // Asynchronous reset active low
```
accuminium: Fix alignment

src/axi_burst_splitter.sv
```diff
@@ -0,0 +1,492 @@
...
+//
+// Copyright and related rights are licensed under the Solderpad Hardware
+// License, Version 0.51 (the "License"); you may not use this file except in
+// compliance with the License.  You may obtain a copy of the License at
+// http://solderpad.org/licenses/SHL-0.51. Unless required by applicable law
+// or agreed to in writing, software, hardware and materials distributed under
+// this License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
```
accuminium: Please attribute Florian, yourself, and me as authors.

src/axi_to_axi_lite.sv
```diff
@@ -9,234 +9,318 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 //
-// Fabian Schuiki <fschuiki@iis.ee.ethz.ch>
-// Andreas Kurth  <akurth@iis.ee.ethz.ch>
```
accuminium: Feel free to add yourself but keep previous authors. They have influenced the current design as well.

src/axi_to_axi_lite.sv
```diff
@@ -9,234 +9,318 @@
...
-    assert(NumPendingRd > 0) else $fatal(1, "Number of pending reads must be greater than 0!");
-    assert(NumPendingWr > 0) else $fatal(1, "Number of pending writes must be greater than 0!");
-  end
-`endif
-// pragma translate_on
 
+  // Assertions
+  // pragma translate_off
+  `ifndef VERILATOR
+  `ifndef SYNTHESIS
```
accuminium: `TARGET_SYNTHESIS`; alternatively omit entirely because `pragma translate_off` should guard

src/axi_to_axi_lite.sv
```diff
@@ -9,234 +9,318 @@
...
+    $fatal(1, $sformatf("Module does not support atomics. Value observed: %0b", slv_req_i.aw.atop));
+  aw_axi_len: assume property( @(posedge clk_i) disable iff (~rst_ni)
+                        slv_req_i.aw_valid |-> (slv_req_i.aw.len == '0)) else
+    $fatal(1, $sformatf("AW request length has to be zero. Value observed: %0b", slv_req_i.aw.len));
+  w_axi_last: assume property( @(posedge clk_i) disable iff (~rst_ni)
+                        slv_req_i.w_valid |-> (slv_req_i.w.last == 1'b1)) else
+    $fatal(1, $sformatf("W last signal has to be one. Value observed: %0b", slv_req_i.w.last));
+  ar_axi_len: assume property( @(posedge clk_i) disable iff (~rst_ni)
+                        slv_req_i.ar_valid |-> (slv_req_i.ar.len == '0)) else
+    $fatal(1, $sformatf("AR request length has to be zero. Value observed: %0b", slv_req_i.ar.len));
```
accuminium: Why `$sformatf` inside `$fatal`? `$fatal` itself supports the formatting you need, right? If so, please simplify.

src/axi_burst_splitter.sv
```diff
@@ -0,0 +1,492 @@
...
+  // Outputs
+  assert property (@(posedge clk_i) mst_req_o.aw_valid |-> mst_req_o.aw.len == '0)
+    else $fatal(1, "AW burst longer than a single beat emitted!");
+  assert property (@(posedge clk_i) mst_req_o.ar_valid |-> mst_req_o.ar.len == '0)
+    else $fatal(1, "AR burst longer than a single beat emitted!");
+  // pragma translate_on
+  `endif
+
+  // TODO: Add FORCE parameter to split non-modifiable bursts with 16 beats or less only when that
+  // parameter is set, to have a standard-compliant alternative?
```
accuminium: What was our conclusion on this TODO item?

src/axi_to_axi_lite.sv
```diff
@@ -9,234 +9,318 @@
...
-`ifndef VERILATOR
-  initial begin: p_assertions
-    assert (AXI_ADDR_WIDTH >= 1) else $fatal(1, "AXI ADDR width must be at least 1!");
-    assert (AXI_DATA_WIDTH >= 1) else $fatal(1, "AXI DATA width must be at least 1!");
-    assert (AXI_ID_WIDTH   >= 1) else $fatal(1, "AXI ID width must be at least 1!");
-    assert (AXI_USER_WIDTH >= 1) else $fatal(1, "AXI USER width must be at least 1!");
+  // Assertions, check params
+  // pragma translate_off
+  `ifndef VERILATOR
+  `ifndef SYNTHESIS
```
accuminium: `TARGET_SYNTHESIS`; alternatively omit entirely because `pragma translate_off` should guard

src/axi_burst_splitter.sv
WRoenninger: We could add the check in the module `axi_burst_splitter_ax_chan`, however I see the problem that letting non modifiable transactions trough would lead to redesign of `axi_burst_splitter_counters` and should be probably handled in its own PR?

accuminium: No, we should not let non-modifiable transactions through without splitting them.

How about we return _Slave Errors_ in this case?

We could implement this by adding a demux at the very front of the burst splitter. If the burst is non-modifiable and shorter than 15 beats (but longer than 1 beat), it goes to the error slave. Also, if it is a wrapping burst, which we currently do not support, it goes to the error slave. Otherwise, it goes to the actual burst splitter as it is currently.

What do you think @WRoenninger @zarubaf ?

WRoenninger: Seems to me the direction we should take.
The assertion also looks at all 4 bits of the `AxCACHE` signal even though only bit `AxCACHE[1]` determines the modifiable aspect.

accuminium: > Seems to me the direction we should take.

:+1: Thanks for the feedback.

> The assertion also looks at all 4 bits of the `AxCACHE` signal even though only bit `AxCACHE[1]` determines the modifiable aspect.

I think it does not, because it does an OR reduction over a bit vector where a single bit is selected by the `axi_pkg::CACHE_MODIFIABLE` mask.

> For decision 2, is the (not incrementing) check necessary?

Yes, AXI specifies "transactions with INCR burst type and length greater than 16 can be converted to multiple small bursts, even if the transaction (...) is Non-modifiable".

accuminium:
I have implemented proper error handling in the burst splitter as discussed above and fixed and improved a few smaller things in the AXI-to-AXI-Lite converter. This branch is now rebased on master and cleaned up. TBs and the synthesis bench are passing, so from my side this is ready to be merged.

accuminium:
@WRoenninger Please review my latest changes (I cannot assign you as reviewer because you created this MR) and let me know if you agree.

WRoenninger:
LGTM
Added default statements in the 3 unique cases as I was getting warnings at simulation start.

PR #36, opened 2020-01-13
fixed compilation and synthesis issues with pkg & intf

da-gazzi:
* intf: import axi_pkg inside of interface bodies - importing outside can lead to compilation issues in ModelSim
* pkg: made tasks automatic - this broke synthesis for me

### DIFF
--- a/src/axi_intf.sv
+++ b/src/axi_intf.sv
@@ -13,7 +13,6 @@
 //
 // This file defines the interfaces we support.
 
-import axi_pkg::*;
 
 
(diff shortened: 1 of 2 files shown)


src/axi_intf.sv
```diff
@@ -13,7 +13,6 @@
 //
 // This file defines the interfaces we support.
 
-import axi_pkg::*;
```
fabianschuiki: Excellent!

zarubaf:
Merging.

PR #43, opened 2020-01-30
AXI DW Conversion

suehtamacv:
This PR proposes two modules, `axi_dw_upsizer` and `axi_dw_downsizer`, for data with conversion, together with a third `axi_dw_converter` module that instantiates one of the previous depending on the relationship between `AxiMstDataWidth` and `AxiSlvDataWidth`.

Given the growing divergence between the `axi_dwc` branch and `master`, instead of rebasing, I considered that it was easier to clean up the work previously hosted there, and create a new branch altogether.


CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## Unreleased
 
 ### Added
+- `axi_dw_converter`: Joins two AXI interfaces, possibly of different data widths.
+  - `axi_dw_downsizer`: Joins a narrow AXI master to a wider AXI slave.
```
bluewww: indendation

src/axi_dw_converter.sv
```diff
@@ -0,0 +1,259 @@
...
+  initial begin
+    assert(slv.AXI_ADDR_WIDTH == mst.AXI_ADDR_WIDTH && slv.AXI_ADDR_WIDTH == AxiAddrWidth) ;
+    assert(slv.AXI_DATA_WIDTH == AxiSlvDataWidth) ;
+    assert(mst.AXI_DATA_WIDTH == AxiMstDataWidth) ;
+    assert(slv.AXI_ID_WIDTH == mst.AXI_ID_WIDTH && slv.AXI_ID_WIDTH == AxiIdWidth) ;
+    assert(slv.AXI_USER_WIDTH == mst.AXI_USER_WIDTH && slv.AXI_USER_WIDTH == AxiUserWidth);
+
+    if (AxiSlvDataWidth != AxiMstDataWidth)
+      assert property (@(posedge clk_i) disable iff (~rst_ni) slv.aw_valid |-> slv.aw_atop[5] == 1'b0)
+      else $fatal(1, "This module does not yet support atomic transactions with an R resp!");
```
bluewww: I don't quite understand the logic of having this in an `initial begin` block. What are the exact semantics here?

src/axi_dw_converter.sv
```diff
@@ -0,0 +1,259 @@
...
+// http://solderpad.org/licenses/SHL-0.51. Unless required by applicable law
+// or agreed to in writing, software, hardware and materials distributed under
+// this License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+// Author: Matheus Cavalcante <matheusd@iis.ee.ethz.ch>
+
+import axi_pkg::*;
+
```
bluewww: move this after `module axi_dw_converter`. Imports should be in the module's scope, not span the whole compilation unit.

src/axi_dw_downsizer.sv
```diff
@@ -0,0 +1,753 @@
...
+// Author: Matheus Cavalcante <matheusd@iis.ee.ethz.ch>
+
+// Description:
+// Data width downsize conversion.
+// Connects a narrow master to a wider slave.
+
+`include "axi/typedef.svh"
+
+import axi_pkg::*;
+
```
bluewww: move this after `module axi_dw_converter`. Imports should be in the module's scope, not span the whole compilation unit.

src/axi_dw_downsizer.sv
```diff
@@ -0,0 +1,753 @@
...
+  // Validate parameters.
+  `ifndef VERILATOR
+  `ifndef SYNTHESIS
+  initial begin: validate_params
+    assert(AxiSlvDataWidth > AxiMstDataWidth)
+    else $fatal(1, "Data downsizer not being used for downsizing.");
+
+    assert (2**AxiIdWidth >= AxiMaxTrans)
+    else $fatal(1, "The outstanding transactions could not be indexed with the given ID bits!");
```
bluewww: Why do you use `inital begin` + `assert` to check parameters? There are elaboration time asserts that do that and they also trigger when you try to synthesis a design violating these constraints (last I checked I think). Tool support might be a bit worse, maybe?

src/axi_dw_upsizer.sv
```diff
@@ -0,0 +1,700 @@
...
+
+// Author: Matheus Cavalcante <matheusd@iis.ee.ethz.ch>
+
+// Description:
+// Data width upsize conversion.
+// Connects a wide master to a narrower slave.
+
+`include "axi/typedef.svh"
+
+import axi_pkg::*;
```
bluewww: same

src/axi_dw_upsizer.sv
```diff
@@ -0,0 +1,700 @@
...
+  // Validate parameters.
+  `ifndef SYNTHESIS
+  `ifndef VERILATOR
+  initial begin: validate_params
+    assert(AxiSlvDataWidth < AxiMstDataWidth)
+    else $fatal(1, "Data upsizer not being used for upsizing.");
+
+    assert (2**AxiIdWidth >= AxiMaxTrans)
+    else $fatal(1, "The outstanding transactions could not be indexed with the given ID bits!");
+  end
```
bluewww: dito. Elaboration time asserts?

test/tb_axi_dw_downsizer.do
```diff
@@ -0,0 +1,4 @@
+add wave -position insertpoint  \
+  sim:/tb_axi_dw_downsizer/i_dw_converter/gen_dw_downsize/i_axi_dw_downsizer/AxiMstDataWidth \
+  sim:/tb_axi_dw_downsizer/i_dw_converter/gen_dw_downsize/i_axi_dw_downsizer/AxiSlvDataWidth \
+  sim:/tb_axi_dw_downsizer/i_dw_converter/gen_dw_downsize/i_axi_dw_downsizer/*
```
bluewww: newline missing

test/tb_axi_dw_upsizer.do
```diff
@@ -0,0 +1,4 @@
+add wave -position insertpoint  \
+  sim:/tb_axi_dw_upsizer/i_dw_converter/gen_dw_upsize/i_axi_dw_upsizer/AxiMstDataWidth \
+  sim:/tb_axi_dw_upsizer/i_dw_converter/gen_dw_upsize/i_axi_dw_upsizer/AxiSlvDataWidth \
+  sim:/tb_axi_dw_upsizer/i_dw_converter/gen_dw_upsize/i_axi_dw_upsizer/*
```
bluewww: newline

test/tb_axi_dw_upsizer.sv
```diff
@@ -0,0 +1,125 @@
...
+// Author        : Matheus Cavalcante <matheusd@iis.ee.ethz.ch>
+// Created       : 09.02.2019
+
+// Copyright (C) 2020 ETH Zurich, University of Bologna
+// All rights reserved.
+
+`include "axi/assign.svh"
+
+module tb_axi_dw_upsizer;
+
```
bluewww: I personally would set `timescale/timeunit`. Xcelium uses per default 1second/1second which is absolutely dumb.

zarubaf: My personal opinion is that the tool argument should override this. I am pretty sure, that when using Xcelium you can override the default timescale argument. Otherwise you get tons of mismatching timescale errors/warnings when trying to simulate.

bluewww: I don't quite understand this argument: This is a top-level testbench, whose modules won't set timescale anywhere so there won't be mismatches. In fact that this testbench already uses hardcoded time values (`700ps` etc.) already implies that there must be a certain `timescale/timeunit` value that is supposed to be used, setting a sane default. Overriding is always possible.

zarubaf: The problem ist that Modelsim will complain about different timescale/time precision units even if you don’t use the module/tomb as top-level. It’s anyway a default timescale for non annotated timings. A comment might be in order.

WRoenninger:
I'd like to point out that since [v0.8.0](https://github.com/pulp-platform/axi/releases/tag/v0.8.0) the default of describing the AXI ports has changed to use structs as defined in [include/axi/typedef.svh](https://github.com/pulp-platform/axi/blob/master/include/axi/typedef.svh) instead of interfaces.

The reason is that not all tools are capable of handling interfaces and that interfaces can lead to unintended behaviour when there is a mismatch in the interface definitions.
An example how the top level module should look like is: [src/axi_cut.sv](https://github.com/pulp-platform/axi/blob/master/src/axi_cut.sv)

I would recommend adding `type` parameters in `axi_dw_converter` and propagate them down the hierarchy over subsequent parametrization. The interface should be in its own `*_intf` module in the same file as `axi_dw_converter`. It should define the required types over the typedef macros and the module can then be used for the testbench.

suehtamacv:
Thanks, everyone for the review. I pushed a few commits handling the issues that were raised.

src/axi_dw_converter.sv
```diff
@@ -0,0 +1,173 @@
...
+    output logic        mst_b_ready_o,
+    output ar_chan_t    mst_ar_o,
+    output logic        mst_ar_valid_o,
+    input  logic        mst_ar_ready_i,
+    input  mst_r_chan_t mst_r_i,
+    input  logic        mst_r_valid_i,
+    output logic        mst_r_ready_o
+  );
+
+  if ($bits(slv_r_o.data) == $bits(mst_r_i.data)) begin: gen_no_dw_conversion
```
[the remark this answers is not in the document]
WRoenninger: Yes I had a problem, where a data type was defined over a `$bits` and this type given as parameter to another module. In this submodule was the `$bits` system function applied on this derived data type parameter. Modelsim complained that chaining of system functions is not allowed in systemverilog and would not simulate. However that problem arose when giving the `data_t` parameter to the function and not the port signals itself as is the case here.
So I think it should be fine, yet should be checked that it does not cause problems.

src/axi_dw_downsizer.sv
```diff
@@ -0,0 +1,599 @@
...
+
+  /****************
+   *  ASSERTIONS  *
+   ****************/
+
+  // Validate parameters.
+  `ifndef SYNTHESIS
+  `ifndef VERILATOR
+  if (2**AxiIdWidth < AxiMaxTrans)
+    $fatal(1, "Cannot index all requested outstanding reads with the given ID bits!");
```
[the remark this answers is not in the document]
suehtamacv: Indeed, I removed this assertion altogether.

src/axi_dw_converter.sv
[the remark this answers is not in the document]
suehtamacv: I am aware of this issue. I ran tests with ModelSim and elaborated the design with DC and had no issues with the `$bits` calls.

suehtamacv:
Yes, @accuminium, these changes correspond to the latest code from the `dwc` branch.

The DW converters do not serialize/pack `WRAP` bursts. These run slowly, but the converter should be compatible with them. Improving their performance should be easy enough. However, the upsizer does *not* support `FIXED` bursts, since these decompose into a myriad of states to handle corner cases in the data width conversion. Should I throw a `$warning()` on these, for the time being? 

This version also includes the commits you did to support atomics transactions. Are they ready to be merged?

src/axi_dw_converter.sv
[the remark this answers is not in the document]
accuminium: Great, thanks for clarifying.

accuminium:
> The DW converters do not serialize/pack `WRAP` bursts. These run slowly, but the converter should be compatible with them. Improving their performance should be easy enough.

Okay, I agree to leave it at this for the moment.

> However, the upsizer does _not_ support `FIXED` bursts, since these decompose into a myriad of states to handle corner cases in the data width conversion. Should I throw a `$warning()` on these, for the time being?

What is the behavior of the upsizer upon a _fixed_ burst?  Raising an `$error()` seems more appropriate to me than a `$warning()`. Am I correctly assuming that replying with a slave error in the upsizer would be too complex?
 
> This version also includes the commits you did to support atomics transactions. Are they ready to be merged?

Yes.

src/axi_dw_downsizer.sv
```diff
@@ -0,0 +1,565 @@
...
+    .NumIn    (AxiMaxReads),
+    .DataType (ar_chan_t  ),
+    .AxiVldRdy(1'b1       ),
+    .ExtPrio  (1'b0       ),
+    .LockIn   (1'b1       )
+  ) i_mst_ar_arb (
+    .clk_i  (clk_i              ),
+    .rst_ni (rst_ni             ),
+    .flush_i(1'b0               ),
+    .rr_i   ('0                 ),
```
bluewww: Doesn't vivado bug when you do `'0` in an instantiation (as in setting only the lowest bit to zero instead of the whole vector). Anyway not sure.

bluewww:
Needs a rebase

src/axi_dw_downsizer.sv
fabianschuiki: Yes it does, that's a good point. However for `'0` that's not a problem, because it gets inferred as a self-determined type of `unsigned logic`, which zero-extends to whatever the actual connection will be. `'1` is problematic, which zero-extends to definitely not what was intended.

suehtamacv: Since `ExtPrio` is set to `1'b0`, the value of the port `rr_i` is ignored. As long as this compiles with Vivado, it should be fine. I will pass the design through it, after submitting another review.

bluewww:
Blocking issue https://github.com/pulp-platform/axi/issues/57

accuminium:
Thanks @suehtamacv, this looks very good now!

Now that the generalized error slave is available on `master` (#55), could you please rebase on `master` and use that error slave?

How about issue #57, has that been addressed?

suehtamacv:
> Thanks @suehtamacv, this looks very good now!
> 
> Now that the generalized error slave is available on `master` (#55), could you please rebase on `master` and use that error slave?

Done. 

> 
> How about issue #57, has that been addressed?

Yes, via a spill register, although further restructuring of the DWC might desirable.

accuminium:
Please implement #63 and let's get this released :slightly_smiling_face:

accuminium:
I am waiting for TBs to pass before merging.

accuminium:
@suehtamacv: The stability property on the R channel is failing in the downsizer TB (line 236 of `axi_intf.sv`, scope `tb_axi_dw_downsizer.axi_master_dv` fails multiple times). Can you investigate?

suehtamacv:
I traced the stability issue of the `R` channel back to the `id_queue`. When a new `AR` transaction arrives, its ID is pushed into the `i_read_id_queue`, to identify which downsizer will handle it. This push operation makes the output port of the `id_queue` to go down, which in turn causes `r_valid` to go down. 

I solved it with 4bafb4a,  by only accepting a new `AR` request when there is no `R` beat on the narrow side. This hinders performance, but it might be acceptable since it goes in the direction of narrow buses (e.g., peripherals). What do you think?

accuminium:
> I solved it with [4bafb4a](https://github.com/pulp-platform/axi/commit/4bafb4a1290be29188cc37ecd421b77304cd7cb4), by only accepting a new `AR` request when there is no `R` beat on the narrow side. This hinders performance, but it might be acceptable since it goes in the direction of narrow buses (e.g., peripherals). What do you think?

In other words, every AR now introduces a "bubble" on the narrow R stream, correct? While I agree with your argument that the downsizer is not used on paths where full throughput is key, this appears to be a significant limitation.

As an alternative, would it be possible to store `idqueue_id` once it is valid (e.g., in a fall-through register) to maintain the stability of the R channel while allowing pushes from ARs?  You would probably have to latch `|idqueue_pop` as well.

suehtamacv:
7ad01b9 removes the ID queue and resolves which downsizer is handling each read request manually, by computing whether the `rid` matches with the `arid` of each downsizer. This gets rid of the bubble on the R stream, and there are no more stability violations of this channel.

accuminium:
> [7ad01b9](https://github.com/pulp-platform/axi/commit/7ad01b9a67dfadd6150ea38bd771fa4f45c44d89) removes the ID queue and resolves which downsizer is handling each read request manually, by computing whether the `rid` matches with the `arid` of each downsizer. This gets rid of the bubble on the R stream, and there are no more stability violations of this channel.

Great, thanks!

Waiting for benches to complete so this can be merged.

PR #44, opened 2020-02-04
Initial support for Riviera-PRO 2019.10

pawelkudlakaldec:


accuminium:
Thanks for your contribution. As it stands, we cannot accept it because Ariane is not a dependency of this repository and this will not change. Can you please describe the issue you are trying to solve and propose another solution?

pawelkudlakaldec:
I agree with previous comment.

PR #45, opened 2020-02-07
Add fix for vivado not correctly extending enum with '1

WRoenninger:
* Fix xbar_latency_e with explicit `CUT_ALL_PORTS`, issue is that vivado incorrectly extends 
`enum logic [9:0] LatencyMode = '1` to 10'b000000001. Workaround: explicit definition
* Add note in [doc/axi_xbar.md](https://github.com/WRoenninger/axi/blob/axi_pkg_vivado_fix/doc/axi_xbar.md) how `LatencyMode` should be configured when connecting two `axi_xbar` in duplex.


WRoenninger:
LGTM

PR #47, opened 2020-02-11
Fix synthesis of `axi_lite_xbar`

accuminium:
Prior to this patch, synthesis in DC failed for `axi_lite_xbar` with `NoMstPorts == 1` due to width mismatches between the `i_axi_aw_decode.idx_o` port and the `dec_aw` signal (and analogously for the AR) and on the `default_mst_port_i`.  The problem was introduced by declaring logic arrays with `[$clog2(Cfg.NoMstPorts)-1:0]`, which evaluated to `[-1:0]` for `Cfg.NoMstPorts == 1`.

### DIFF
--- a/src/axi_lite_xbar.sv
+++ b/src/axi_lite_xbar.sv
@@ -26,18 +26,19 @@ module axi_lite_xbar #(
   parameter type  r_chan_t          = logic,
   parameter type     req_t          = logic,
   parameter type    resp_t          = logic,
-  parameter type    rule_t          = axi_pkg::xbar_rule_64_t
+  parameter type    rule_t          = axi_pkg::xbar_rule_64_t,
+  localparam int unsigned MstIdxWidth = (Cfg.NoMstPorts > 32'd1) ? $clog2(Cfg.NoMstPorts) : 32'd1
 ) (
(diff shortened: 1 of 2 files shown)


src/axi_lite_xbar.sv
```diff
@@ -26,18 +26,19 @@ module axi_lite_xbar #(
   parameter type  r_chan_t          = logic,
   parameter type     req_t          = logic,
   parameter type    resp_t          = logic,
-  parameter type    rule_t          = axi_pkg::xbar_rule_64_t
+  parameter type    rule_t          = axi_pkg::xbar_rule_64_t,
+  localparam int unsigned MstIdxWidth = (Cfg.NoMstPorts > 32'd1) ? $clog2(Cfg.NoMstPorts) : 32'd1
```
WRoenninger: I would change this to use a parameter as not all tools are able to handle localparams in parameter list.
```suggestion
  // DEPENDENT PARAMETERS, DO NOT OVERWRITE!
  parameter int unsigned MstIdxWidth = (Cfg.NoMstPorts > 32'd1) ? $clog2(Cfg.NoMstPorts) : 32'd1
```

PR #48, opened 2020-02-12
Fix synthesis of `axi_demux_wrap`, `axi_lite_demux_intf`, and `axi_lite_mux`

accuminium:


accuminium:
> Should we not also change the module name of `axi_demux_wrap` to `axi_demux_intf`?

Yes, but let's do that in a separate MR.  Rationale: first fix what is backward-compatible.

> Rest of the changes LGTM.

Thanks!

PR #55, opened 2020-02-18
Generalize Error Slave

accuminium:
This is a break-out from #43.  `axi_decerr_slv` has been replaced by a more generic `axi_err_slv`, which takes the kind of error as parameter.


accuminium:
Additionally, I would like to drop support for ATOPs from `axi_err_slv`.  If this slave is required to handle ATOPs, you now have to place an `axi_atop_filter` in front of it. This prevents wasting resources in cases when there can be no ATOPs, such as our `axi_lite_xbar`, where the `axi_err_slv` is also instantiated.

Do you agree, @WRoenninger and @zarubaf ?

src/axi_err_slv.sv
```diff
@@ -10,15 +10,17 @@
...
-  parameter int unsigned MaxTrans    = 1     // Maximum # of accepted transactions before stalling
+// AXI Error Slave: This module always responds with an AXI error for transactions that are sent to
+// it.  This module does NOT support ATOPs; you must place an `axi_atop_filter` in front of it if
+// you need it to handle ATOPs.
+
+module axi_err_slv #(
+  parameter int unsigned    AxiIdWidth  = 0,                    // AXI ID Width
+  parameter type            req_t       = logic,                // AXI 4 request struct, with atop field
+  parameter type            resp_t      = logic,                // AXI 4 response struct
+  parameter bit             FallThrough = 1'b1,                 // When enabled: in cycle transaction, long paths
```
WRoenninger: I think we should drop the `FallThrough` parameter to break combinational paths to the responses

src/axi_err_slv.sv
```diff
@@ -55,29 +54,12 @@ module axi_decerr_slv #(
...
   // Write Transactions
   //--------------------------------------
   // push, when there is room in the fifo
-  assign w_fifo_push          = decerr_req.aw_valid & ~w_fifo_full;
-  assign decerr_resp.aw_ready = ~w_fifo_full;
+  assign w_fifo_push          = slv_req_i.aw_valid & ~w_fifo_full;
+  assign slv_resp_o.aw_ready  = ~w_fifo_full;
 
   fifo_v3 #(
     .FALL_THROUGH ( FallThrough ),
```
WRoenninger: Have one cycle latency at least, and update whitespace accordingly
```suggestion               
    .FALL_THROUGH ( 1'b0        ),
```

src/axi_err_slv.sv
```diff
@@ -10,15 +10,17 @@
...
+// it.  This module does NOT support ATOPs; you must place an `axi_atop_filter` in front of it if
+// you need it to handle ATOPs.
+
+module axi_err_slv #(
+  parameter int unsigned    AxiIdWidth  = 0,                    // AXI ID Width
+  parameter type            req_t       = logic,                // AXI 4 request struct, with atop field
+  parameter type            resp_t      = logic,                // AXI 4 response struct
+  parameter bit             FallThrough = 1'b1,                 // When enabled: in cycle transaction, long paths
+  parameter axi_pkg::resp_t Resp        = axi_pkg::RESP_DECERR, // Error generated by this slave.
+  parameter int unsigned    MaxTrans    = 1                     // Maximum # of accepted transactions before stalling
```
WRoenninger: Additionally, would it be more efficient to use spill_registers instead of the FIFOs?
This module answers with an error so each master which makes a request here anyway does something not efficient, so the support of multiple concurrent transactions would not be required.

src/axi_err_slv.sv
accuminium: I agree that we should disable falling through from W to B channel (i.e., in the `i_b_fifo`), but falling through from AW to W channel should be fine, no? The error slave should be able to absorb the burst that will cause an error response as soon as possible to prevent stalls on upstream buses.

src/axi_err_slv.sv
accuminium: I agree with dropping this parameter.

src/axi_err_slv.sv
accuminium: No, in the interest of freeing the upstream bus, the error slave should be configurable to drain multiple requests without having completed its responses.

accuminium:
TBs and the synthesis bench are passing, so I am merging this.

PR #56, opened 2020-02-18
Small improvements to atop_filter, demux, id_prepend, and xbar

accuminium:
### Added
- `axi_xbar_intf`: Add interface variant of crossbar.

### Fixed
- `axi_atop_filter`: Fix ModelSim warnings by adding `default` statement.  The signal in the `case`   has a single bit, and both values were correctly handled in synthesis.  However, when starting   simulation, the signal has an undefined value, and ModelSim threw warnings that this violated the   `unique` condition.
- `axi_demux`: Move `typedef` outside `generate` for compatibility with VCS.
- `axi_id_prepend`:
  - Fix text of some assertion messages.
  - Fix case of prepending a single-bit ID.

### DIFF
--- a/src/axi_demux.sv
+++ b/src/axi_demux.sv
@@ -51,24 +51,25 @@ module axi_demux #(
 
   localparam int unsigned IdCounterWidth = $clog2(MaxTrans);
 
+  //--------------------------------------
+  // Typedefs for the FIFOs / Queues
+  //--------------------------------------
+  typedef logic [AxiIdWidth-1:0] axi_id_t;
+  typedef struct packed {
+    aw_chan_t aw_chan;
+    select_t  aw_select;
+  } aw_chan_select_t;
+  typedef struct packed {
+    ar_chan_t ar_chan;
+    select_t  ar_select;
+  } ar_chan_select_t;
+
   // pass through if only one master port
   if (NoMstPorts == 32'h1) begin : gen_no_demux
     assign mst_reqs_o[0] = slv_req_i;
     assign slv_resp_o    = mst_resps_i;
   // other non degenerate cases
   end else begin : gen_demux
-    //--------------------------------------
-    // Typedefs for the Fifo's / Queues
-    //--------------------------------------
-    typedef logic [AxiIdWidth-1:0] axi_id_t;
-    typedef struct packed {
-      aw_chan_t aw_chan;
-      select_t  aw_select;
-    } aw_chan_select_t;
-    typedef struct packed {
-      ar_chan_t ar_chan;
-      select_t  ar_select;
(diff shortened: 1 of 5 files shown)


src/axi_id_prepend.sv
```diff
@@ -76,17 +76,20 @@ module axi_id_prepend #(
...
 
   // prepend the ID
-  always_comb begin
-    for (int unsigned i = 0; i < NoBus; i++) begin : gen_id_prepend
-      mst_aw_chans_o[i]                                   = slv_aw_chans_i[i];
-      mst_aw_chans_o[i].id[AxiIdWidthSlvPort+:PreIdWidth] = pre_id_i;
-      mst_ar_chans_o[i]                                   = slv_ar_chans_i[i];
-      mst_ar_chans_o[i].id[AxiIdWidthSlvPort+:PreIdWidth] = pre_id_i;
-      // The ID is in the highest bits of the struct, so an assignment from a channel with a wide ID
-      // to a channel with a shorter ID correctly cuts the prepended ID.
```
zarubaf: We can keep the comment, right? It still applies, doesn't it?

src/axi_demux.sv
```diff
@@ -60,15 +70,6 @@ module axi_demux #(
     //--------------------------------------
     // Typedefs for the Fifo's / Queues
     //--------------------------------------
```
zarubaf: Should we move the comment to the new code section above? Or delete it?

WRoenninger: I would move it to the typedefs

src/axi_id_prepend.sv
accuminium: Thanks, I forgot to move that comment. Fixed.

src/axi_demux.sv
accuminium: Indeed, I forgot to move that comment. Fixed.

src/axi_xbar.sv
```diff
@@ -220,3 +220,90 @@ module axi_xbar #(
...
   // pragma translate_on
 endmodule
+
+`include "axi/assign.svh"
+`include "axi/typedef.svh"
+
+module axi_xbar_intf #(
+  parameter int unsigned AXI_USER_WIDTH = 0,
+  parameter axi_pkg::xbar_cfg_t Cfg = '0,
+  parameter type rule_t             = axi_pkg::xbar_rule_64_t
```
WRoenninger: Should we align the parameters?

accuminium: Fixed, thanks.

accuminium:
Thanks for your reviews!

TBs and the synthesis bench are passing, so I am rebasing and merging this.

PR #59, opened 2020-02-20
AXI4-Lite to APB4 Bridge: Change APB request port

WRoenninger:
* Change that the port `apb_req_o` is now an array with `NoApbSlaves` number of fields.
* The module now asserts that the width of the `psel` signal in the APB4 request struct is `1'b1`.
* The `apb_req_o` port is now default `'0` and only goes active for a correctly decoded slave on the respective port array field when an APB transaction takes place.

### DIFF
--- a/src/axi_lite_to_apb.sv
+++ b/src/axi_lite_to_apb.sv
@@ -23,15 +23,14 @@
 //   typedef struct packed {
 //     addr_t paddr;   // same as AXI4-Lite
 //     prot_t pprot;   // same as AXI4-Lite, specification is the same
-//     sel_t  psel;    // onehot, one psel line per connected APB4 slave
+//     logic  psel;    // each APB4 slave has its own single-bit psel
 //     logic  penable; // enable signal shows second APB4 cycle
 //     logic  pwrite;  // write enable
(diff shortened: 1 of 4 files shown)


src/axi_lite_to_apb.sv
```diff
@@ -23,15 +23,14 @@
...
 //     logic  penable; // enable signal shows second APB4 cycle
 //     logic  pwrite;  // write enable
 //     data_t pwdata;  // write data, comes from W channel
 //     strb_t pstrb;   // write strb, comes from W channel
 //   } apb_req_t;
-// To connect one APB4 slave the corresponding `psel` signal only asserts, when the address
-// decoding matches. To connect to a single APB4 slave connect each of the `apb_req_o` signals,
-// however only the corresponding `psel` bit.
+// For every APB4 slave, the `psel` field is only asserted when the decoded address matches that
+// slave.  If `psel` is deasserted, the value of the other fields may be undefined.
```
accuminium: I found this confusing and rewrote it.  @WRoenninger: please confirm whether you agree.

WRoenninger: LGTM

PR #61, opened 2020-02-28
axi_demux: Add `unique` to `case` distinction

accuminium:
The `case` statement controlling the counters had not been specified `unique` even though it qualified for it.  This has been fixed.


accuminium:
Thanks!

PR #64, opened 2020-03-02
axi_atop_filter: Fix W handshake

WRoenninger:
When the connected slave module wants both AW and W to be valid before asserting thier readys, `axi_atop_filter` was stalling forever and did not let the W valid through.
Added a second condition for when AW is valid and non ATOP then the W handshake gets connected.


accuminium:
I refactored your change a bit and cleaned up the fixup.  Behavioral TBs and synthesis bench are passing. Merging.

PR #65, opened 2020-03-05
AXI4-Lite to APB4 Bridge: Update interface version to match changed req ports 

SamuelRiedel:
- Change the interface version of the `axi_lite_to_apb` version to match the changes done in #59

### DIFF
--- a/src/axi_lite_to_apb.sv
+++ b/src/axi_lite_to_apb.sv
@@ -358,14 +358,16 @@ module axi_lite_to_apb_intf #(
   // APB Slave Address Map
   input  rule_t [NoRules-1:0]     addr_map_i
 );
+  localparam int unsigned SelIdxWidth = NoApbSlaves > 1 ? $clog2(NoApbSlaves) : 1;
+
   typedef struct packed {
-     addr_t          paddr;   // same as AXI4-Lite
-     axi_pkg::prot_t pprot;   // same as AXI4-Lite, specification is the same
-     sel_t           psel;    // onehot, one psel line per connected APB4 slave
-     logic           penable; // enable signal shows second APB4 cycle
-     logic           pwrite;  // write enable
-     data_t          pwdata;  // write data, comes from W channel
-     strb_t          pstrb;   // write strb, comes from W channel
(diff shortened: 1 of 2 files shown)


SamuelRiedel:
@WRoenninger, @accuminium  Feedback is appreciated.

src/axi_lite_to_apb.sv
```diff
@@ -358,14 +358,16 @@ module axi_lite_to_apb_intf #(
   // APB Slave Address Map
   input  rule_t [NoRules-1:0]     addr_map_i
 );
+  localparam int unsigned SelBinWidth = NoApbSlaves > 1 ? $clog2(NoApbSlaves) : 1;
```
accuminium: Please [use `ALL_CAPS` to name constants](https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#construct-naming).

SamuelRiedel: I think the style guide is a bit ambiguous here. In the [Constants](https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#constants) section, it says:

> The preferred naming convention for all immutable constants is to use `ALL_CAPS`, but there are times when the use of `UpperCamelCase` might be considered more natural.

and later it specifies to use _UpperCamelCase_ for derived localparams. So I tried to be consistent with the non-interface version: [`localparam int unsigned SelIdxWidth = (NoApbSlaves > 32'd1) ? $clog2(NoApbSlaves) : 32'd1;`](https://github.com/pulp-platform/axi/blob/b8c10cfa26ede50d2546e757e6f57085fe4c9a65/src/axi_lite_to_apb.sv#L71)

But to be truly consistent I think I should rename the variable to `SelIdxWidth` but keep the _UpperCamelCase_. What do you think?

accuminium: You are right, `UpperCamelCase` it is because this is a _derived localparam_. Thanks!

I agree with the rename to `SelIdxWidth`.

SamuelRiedel: :+1: Done.

PR #66, opened 2020-03-05
axi_demux: Fix case when `MaxTrans` is 1

SamuelRiedel:
In case the `MaxTrans` parameter was set to 1, the `IdCounterWidth` was calculated to be 0.


accuminium:
Rebased to fix conflict in Changelog. Merging.

PR #69, opened 2020-03-09
Axi dwc fix typedef

WRoenninger:
This includes fixes which make the module compilable and simulatable by Verilator:
* Changed the position of struct definitions to be outside the generate constructs.
* Renamed `*state_t` to `*state_e`.
* Removed `import axi_pkg::*;` and made calls to the package explicit. Problem is that otherwise the definitions from the package are polluting the namespace globally for all modules compiled by Verilator after this one.


src/axi_dw_downsizer.sv
```diff
@@ -45,8 +45,6 @@ module axi_dw_downsizer #(
     input  axi_mst_resp_t mst_resp_i
   );
 
-  import axi_pkg::*;
-
```
accuminium: @WRoenninger : Does Verilator make the symbols imported here visible beyond the scope of this module?

WRoenninger: Added the explicit import of `aligned_addr`.
However I think for the other imported signal types it is better to leave the explicit package call, as it makes it clear where the type came from.

accuminium: Okay.

PR #73, opened 2020-03-13
axi_lite_to_axi: Add Data Width parameter to fix synthesis in Xcelium

accuminium:
Reportedly, Xcelium (version unknown) cannot handle hierarchical expressions (e.g., `mst_req_o.w.data`) as argument to the `$bits()` function. I propose to fix this in this PR by adding a parameter for the data width.  The disadvantage is that this is a backwards-incompatible "API" change. I am open to backwards-compatible suggestions, of course.

As we currently do not have regression tests for Xcelium, -- at the moment we cannot verify every version of every tool with its own flavor of SystemVerilog -- I suggest to fix similar issues reactively on a per-case basis.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 ### Changed
+- `axi_lite_to_axi`: Add mandatory `AxiDataWidth` parameter to enable fix mentioned below.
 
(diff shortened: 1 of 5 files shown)


accuminium:
@zarubaf: Do you also agree with this fix and the proposed way of handling tool-specific bugs?

accuminium:
Thanks for the reviews!

Rebased on `master`, TBs passing. Merging.

PR #74, opened 2020-03-13
axi_intf: Improve Verilator compatibility

### LINKED ISSUE #68: Verilator compile issues
I am working on simulating the AXI4 code in Verilator and are running into a problem with the parameter types of the modules. The issue is the `parameter type resp_t = logic` definition in al modules. However renaming the definitions to `parameter type rsp_t = logic` seems to not trigger the issue. I think the issue could be that in `axi_pkg` the parameter `resp_t` is already defined and part of the response struct which is the parameter which has this problem. 
Should I do a fix for all modules, this will however break all other projects which are using the new AXI4 infrastructure currently. 

Related to Verilator: In #43 there is in file `src/axi_dw_upsizer.sv` line 321 a struct definition for `r_req_d` and `r_req_q` inside a generate. Verilator complains about multiple type definitions.
I was able to fix this by declaring a separate typedef outside the generate and defining the signals separate in the generate:

```systemverilog
typedef struct packed {
      ar_chan_t ar         ;
      logic ar_valid       ;
      logic ar_throw_error ;
      slv_r_chan_t r       ;
      logic r_valid        ;
      burst_len_t burst_len;
      size_t orig_ar_size  ;
} r_req_t;

// code

for (genvar ...
  r_req_t r_req_d, r_req_q;
end
```

accuminium:
Closes #68.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 - Remove whitespace in and semicolon after macro calls.
+- `axi_intf`: Improve Verilator compatibility by disabling unsupported assertions.
 
 
 ## 0.16.0 - 2020-03-11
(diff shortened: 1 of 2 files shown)


accuminium:
TBs passing. Merging.

PR #76, opened 2020-03-16
Error Slave Response Value

WRoenninger:
This adds to parameter to `axi_err_slv` to help with debugging on a running system:
* `RespWidth`: Response width of the value on the R channel. Does not have to match `AxiDataWidth`.
* `RespData`: The value returned on `slv_resp_o.r.data`, zero extended or truncated, depends on `AxiDataWidth` in relation to `RespWidth`.
Should I propagate the parameter to the outside of modules which instantiate `axi_err_slv`, namely `axi_xbar`, `axi_lite_xbar` and `axi_burst_splitter?


accuminium:
> Should I propagate the parameter to the outside of modules which instantiate `axi_err_slv`, namely `axi_xbar`, `axi_lite_xbar` and `axi_burst_splitter?

No, I think this would be "feature creep".  Let's add such things when we have the need for them.

accuminium:
TBs are running, I will merge once they have all passed.

PR #79, opened 2020-03-20
AXI4-Lite Registers

WRoenninger:
This adds the module `axi_lite_regs`, with test_bench and synth_bench.
Features:
* Register array address aligned to the data width of the AXI4-Lite channel.
* Optional: Read-only mapping per register index, then the init-value gets returned.
* Optional: `PROT` signal functionality.


test/tb_axi_lite_regs.sv
```diff
@@ -0,0 +1,172 @@
...
+
+  //-----------------------------------
+  // DUT
+  //-----------------------------------
+  axi_lite_regs_intf #(
+    .NUM_AXI_REGS   ( NumAxiRegs   ),
+    .AXI_ADDR_WIDTH ( AxiAddrWidth ),
+    .AXI_DATA_WIDTH ( AxiDataWidth ),
+    .REG_DATA_WIDTH ( RegDataWidth ),
+    .READ_ONLY      ( ReadOnly     )
```
accuminium, line 163: Why are not all DUT parameters forwarded to the DUT?

test/tb_axi_lite_regs.sv
```diff
@@ -0,0 +1,172 @@
...
+module tb_axi_lite_regs #(
+  // register configuration
+  parameter int unsigned           NumAxiRegs   = 32'd32,
+  parameter int unsigned           RegDataWidth = 32'd27,
+  parameter logic [NumAxiRegs-1:0] ReadOnly     = {NumAxiRegs{1'b0}},
+  // AXI configuration
+  parameter int unsigned           AxiAddrWidth = 32'd32,    // Axi Address Width
+  parameter int unsigned           AxiDataWidth = 32'd32,    // Axi Data Width
+  parameter bit                    PrivProtOnly = 1'b0,
+  parameter bit                    SecuProtOnly = 1'b0,
```
accuminium, line 26: Can we add basic tests for the `ReadOnly`, `PrivProtOnly`, and `SecuProtOnly` features?

src/axi_lite_regs.sv
```diff
@@ -0,0 +1,334 @@
...
+  ) i_r_spill_register (
+    .clk_i,
+    .rst_ni,
+    .valid_i ( r_valid ),
+    .ready_o ( r_ready ),
+    .data_i  ( r_chan  ),
+    .valid_o ( axi_resp_o.r_valid ),
+    .ready_i ( axi_req_i.r_ready  ),
+    .data_o  ( axi_resp_o.r       )
+  );
```
accuminium, line 236: Please add a comment to describe why these spill registers are needed. I wonder if a `fall_through_register` is not sufficient to have ready-when-idle upstream?

Bender.yml
```diff
@@ -38,6 +38,7 @@ sources:
   - src/axi_lite_join.sv
   - src/axi_lite_mailbox.sv
   - src/axi_lite_mux.sv
+  - src/axi_lite_regs.sv
```
accuminium, line 41: Please also update `src_files.yml`.

src/axi_lite_regs.sv
```diff
@@ -0,0 +1,334 @@
...
+    // Control
+    if (axi_req_i.aw_valid && axi_req_i.w_valid) begin
+      if (aw_dec_valid && aw_prot_ok && !ReadOnly[aw_idx]) begin
+        b_chan.resp = axi_pkg::RESP_OKAY;
+        for (int unsigned i = 0; i < RegDataWidth; i++) begin
+          if (axi_req_i.w.strb[i/8]) begin
+            reg_d[aw_idx][i] = axi_req_i.w.data[i];
+          end
+        end
+        reg_update[aw_idx] = 1'b1;
```
accuminium, line 139: Why `reg_update` and a FFLARN instead of a simpler FFARN? You already change `reg_d` only for a specific `[aw_idx]`?

src/axi_lite_regs.sv
```diff
@@ -0,0 +1,334 @@
...
+                      (SecuProtOnly ? axi_req_i.aw.prot[1] : 1'b1);
+  always_comb begin
+    // default assignments
+    reg_d               = reg_q;
+    reg_update          = '0;
+    // Channel handshake
+    axi_resp_o.aw_ready = 1'b0;
+    axi_resp_o.w_ready  = 1'b0;
+    // Response
+    b_chan              = '0;
```
accuminium, line 128:
```suggestion
    b_chan              = '{default: '0};
```
Some tools complain when assigning an array literal to a packed struct.

src/axi_lite_regs.sv
```diff
@@ -0,0 +1,334 @@
...
+  logic         ar_prot_ok;
+  assign ar_prot_ok = (PrivProtOnly ? axi_req_i.ar.prot[0] : 1'b1) &
+                      (SecuProtOnly ? axi_req_i.ar.prot[1] : 1'b1);
+  assign r_chan = '{
+    data: (ar_dec_valid && ar_prot_ok) ?
+              ( ReadOnly[ar_idx] ?
+                  axi_data_t'(reg_init_i[ar_idx]) :
+                  axi_data_t'(reg_q[ar_idx])
+              ) :
+              axi_data_t'(32'hDEADBEEF),
```
accuminium, line 165: `DEADBEEF` is overused :-) Please find a [less frequently used (unoffending) alternative](https://en.wikipedia.org/wiki/Hexspeak).

src/axi_lite_regs.sv
```diff
@@ -0,0 +1,334 @@
...
+
+  // Read logic
+  reg_idx_t     ar_idx;
+  logic         ar_dec_valid;
+  r_chan_lite_t r_chan;
+  logic         r_valid,      r_ready;
+  logic         ar_prot_ok;
+  assign ar_prot_ok = (PrivProtOnly ? axi_req_i.ar.prot[0] : 1'b1) &
+                      (SecuProtOnly ? axi_req_i.ar.prot[1] : 1'b1);
+  assign r_chan = '{
```
accuminium, line 159: I think this assignment is complex enough to be worth an `always_comb`.

src/axi_lite_regs.sv
WRoenninger: The responses get generated in the same cycle as a request gets accepted. The spill registers are needed to ensure no comb paths between slave port inputs and outputs.

src/axi_lite_regs.sv
```diff
@@ -0,0 +1,339 @@
...
+              axi_pkg::RESP_SLVERR,
+    default : '0
+  };
+  assign r_valid             = axi_req_i.ar_valid; // to spill register
+  assign axi_resp_o.ar_ready = r_ready;            // from spill register
+
+  for (genvar i = 0; i < NumAxiRegs; i++) begin : gen_rw_regs
+    assign reg_q_o[i] = ReadOnly[i] ? reg_init_i[i] : reg_q[i];
+  end
+  `FFLARN(reg_q, reg_d, reg_update, reg_init_i, clk_i, rst_ni)
```
accuminium, line 166: Please consider that [reset values driven from signals are problematic](https://github.com/pulp-platform/apb/pull/3) and rewrite the initialization accordingly.

src/axi_lite_regs.sv
WRoenninger: `reg_update` in this case should fall under the second point?
 
> Long story short, my personal opinion: 
> - Code readability has priority, which IMHO is inhibited in this case.
> - Load-enables for larger structures and flipflops which occasionally latch data.
> 
> Originally posted by @zarubaf in #93

src/axi_lite_regs.sv
WRoenninger: I thought about adding some sort of `reg_set` functionality. So that the internal registers can be loaded/updated with the `reg_init` signal. In the first cycle after reset all register would get set to the init value. 
Would be useful to have some writable registers also from logic than only from the AXI-Lite side. Use case: Switch for trigger some functionality (LLC flush) that deasserts itself when done.

src/axi_lite_regs.sv
WRoenninger: Used `BA5E1E55` (baseless) instead.

src/axi_lite_regs.sv
accuminium: I fully agree that there should be an input signal, `reg_init_i`, that defines the value of the registers after reset. However, I disagree with the current implementation, because it means that the reset value is defined by an input signal, which can lead to timing closure problems (i.e., `reg_init_i` needs to adhere to timing constraints w.r.t. `rst_ni`). Instead, I suggest to adapt the linked approach, where `reg_d` is assigned to `reg_init_i` in the first cycle *after* `rst_ni` is high.

Your second point, being able to set the value of the registers through another stream, is orthogonal. It is also worth considering, but let's solve one issue at a time.

WRoenninger: Should I keep the reg_init separate or combine it with a loading functionality?
Something like this if there is a separate load input:
```systemverilog
    // Handle all non AXI register loads.
    // External from load signal and initialization load after reset.
    for (int unsigned i = 0; i < NumAxiRegs; i++) begin
      if ((reg_load_i[i] || !has_reset_q) && !ReadOnly[i]) begin
        reg_d[i]      = reg_d_i[i];
        reg_update[i] = 1'b1;
      end
    end
```

WRoenninger: For the not AXI triggered register load:
Which load strategy would be the most preferable?
* Load is stronger than AXI write. -> AXI write responds with error when load in same cycle.
* AXI is stronger than load. -> AXI write gets executed and load is ignored.
* AXI write stalls as long as the load is asserted. -> Write executed when load is 0.

accuminium: > Should I keep the reg_init separate or combine it with a loading functionality?

Separate because otherwise the user has to multiplex initialization and load values.

> Something like this if there is a separate load input:

When you separate loading, the condition would need to be
```sv
if (!ReadOnly[i] && has_reset_q && reg_load_i[i]) begin
```
right?

> Which load strategy would be the most preferable?

AXI write stalls as long as the load is asserted. -> Write executed when load is 0.

But make sure to state this in the port description of `reg_load_i`.

src/axi_lite_regs.sv
accuminium: Okay, agreed.

WRoenninger:
I had to add the AXI4-Lite field `aw_prot` and `ar_prot` to the interface modport, as I am using these fields in `tb_axi_lite_regs`.
I updated the AXI4-Lite testing infrastructure accordingly. I think I caught all instances, where the new fields made an impact on the code. 
Added all influences of these additional signals to the `CHANGELOG`.

src/axi_lite_regs.sv
```diff
@@ -0,0 +1,480 @@
...
+  /// AXI4-Lite slave port request struct, bundles all AXI4-Lite signals from the master.
+  /// This module will perform the write transaction when `axi_req_i.aw_valid` and
+  /// `axi_req_i.w_valid` are both asserted.
+  input  req_lite_t               axi_req_i,
+  /// AXI4-Lite slave port response struct, bundles all AXI4-Lite signals to the master.
+  output resp_lite_t              axi_resp_o,
+  /// Flag to tell logic that the AXI is writing these bytes this cycle. The new data
+  /// where `wr_active[byte_index] && AxiReadOnly[byte_index] == 1'b0` is available in the
+  /// next cycle at `reg_q_o[byte_index]`. This signal is directly generated from the signal
+  /// axi_req_i.w.strb and is asserted regardless of the value of `AxiReadOnly`.
```
accuminium, line 103: Why does this signal not take `AxiReadOnly` into consideration?

src/axi_lite_regs.sv
```diff
@@ -0,0 +1,480 @@
...
+
+    // Handle load from AXI write.
+    // `b_ready` is allowed to be a condition as it comes from a spill register.
+    if (axi_req_i.aw_valid && axi_req_i.w_valid && b_ready) begin
+      // The write can be performed when these conditions are true:
+      // - AW decode is valid.
+      // - `axi_req_i.aw.prot` has the right value.
+      if (aw_dec_valid && aw_prot_ok) begin
+        // Stall write as long as any direct load is going on in the current chunk.
+        if (!chunk_loaded[aw_chunk_idx]) begin
```
accuminium, line 233: This is problematic. Imaging one read-only byte that is constantly being loaded adjacent one regular (read-write) byte. When both bytes fall within the same chunk, writing the regular byte stalls the AXI4-Lite port forever.

Instead, I suggest to stall only writes whose strobe targets a non-read-only byte while that byte is being loaded.

accuminium:
Very nice, this is getting mature.

I have just overhauled the [specification of the module](https://pulp-platform.github.io/axi/axi_lite_regs/module.axi_lite_regs.html) with two goals: (1) simplify descriptions and minimize redundancy and (2) remove notion of *chunks* (i.e., bundle of bytes in the same data beat) from the specification. Having *chunks* in the specification causes the behavior of this module to depend on the data width of the AXI4-Lite port, which, as discussed, we absolutely want to avoid.

As [mentioned above](https://github.com/pulp-platform/axi/pull/79#discussion_r419606468), there is one aspect left where behavior currently depends on *chunks*, but I think that can be fixed easily.

Please review [my changes to the specification](https://github.com/pulp-platform/axi/pull/79/commits/0fdda5d046c955478af21dc8015b0c01e3bb8e60). If you agree, let's freeze it and get it verified and merged.

src/axi_lite_regs.sv
WRoenninger: So that surrounding logic knows attempted writes to these bytes to trigger an error interrupt for example.

accuminium: Okay, makes sense. I have added that reason to the docstring.

src/axi_lite_regs.sv
WRoenninger: fbf45f8 changes the stall behavior to the suggestion.

accuminium: :+1: thanks

src/axi_lite_regs.sv
```diff
@@ -0,0 +1,473 @@
...
+  localparam int unsigned AxiStrbWidth  = AxiDataWidth / 32'd8;
+  localparam int unsigned NumChunks     = cf_math_pkg::ceil_div(RegNumBytes, AxiStrbWidth);
+  localparam int unsigned ChunkIdxWidth = (NumChunks > 32'd1) ? $clog2(NumChunks) : 32'd1;
+  // Type of the index to identify a specific register chunk.
+  typedef logic [ChunkIdxWidth-1:0] chunk_idx_t;
+
+  // Find out how many bits of the address are applicable for this module.
+  // Look at the `AddrWidth` number of LSBs to calculate the multiplexer index of the AXI.
+  localparam int unsigned AddrWidth = (RegNumBytes > 32'd1) ? $clog2(RegNumBytes) : 32'd1;
+  typedef logic [AddrWidth-1:0] addr_t;
```
accuminium, line 148: `addr_t` needs to be one bit wider, because `end_addr` of the address decoder is not inclusive. Currently, [`end_addr` overflows for the last chunk](https://github.com/pulp-platform/axi/pull/79/files#diff-0746c8b0c1a5a81fdad737eea8b46101R161) if the address space is fully used.

WRoenninger:
Stall flags on write are now one bit, updated `synth_bench`.

accuminium:
Any final comments, @zarubaf?

accuminium:
CI status

README.md
```diff
@@ -28,6 +28,7 @@ This is the implementation of the AMBA AXI protocol developed as part of the PUL
 | [`axi_lite_join`](src/axi_lite_join.sv)              | A connector that joins two AXI-Lite interfaces.                                                   |                                |
 | [`axi_lite_mailbox`](src/axi_lite_mailbox.sv)        | A AXI4-Lite Mailbox with two slave ports and usage triggered irq.                                 | [Doc](doc/axi_lite_mailbox.md) |
 | [`axi_lite_mux`](src/axi_lite_mux.sv)                | Multiplexes AXI4-Lite slave ports down to one master port.                                        | [Doc](doc/axi_lite_mux.md)     |
+| [`axi_lite_regs`](src/axi_lite_regs.sv)              | AXI4-Lite registers with optional read-only and protection features.                              |                                |
```
zarubaf, line 31: Should we maybe link to the auto-generated documentation maybe?

src/axi_lite_regs.sv
```diff
@@ -0,0 +1,479 @@
...
+///
+/// ### Exposing Constants
+///
+/// To make a byte with constant value (e.g., implemented as LUT instead of FF after synthesis)
+/// readable from the AXI4-Lite port:
+/// - Make the byte read-only from the AXI4-Lite port by setting its `AxiReadOnly` bit to `1`.
+/// - Disable loading the byte from logic by driving its `reg_load_i` bit to `0`.
+/// - Define the value of the byte by setting its `RegRstVal` entry.
+///
+/// ## Protection
```
zarubaf, line 53:
```suggestion
/// ## Protection

```

src/axi_lite_regs.sv
```diff
@@ -0,0 +1,479 @@
...
+/// This module can be configured to only allow *privileged* and/or *secure* accesses by setting the
+/// `PrivProtOnly` and/or `SecuProtOnly` parameter, respectively.
```
zarubaf, line 55: Maybe add a sentence that this is from the AXI perspective and that the "Protection" and "Secure" bits are defined in the spec.

src/axi_test.sv
```diff
@@ -1435,25 +1451,25 @@ package axi_test;
     endtask
 
     // write data to a specific address
-    task automatic write(input addr_t w_addr, input data_t w_data, input strb_t w_strb,
-                         output axi_pkg::resp_t b_resp);
-      $display("%0t %s> Write to ADDR: %h, DATA: %h, STRB: %h",
-          $time(), this.name, w_addr, w_data, w_strb);
+    task automatic write(input addr_t w_addr, input prot_t w_prot, input data_t w_data,
+                         input strb_t w_strb, output axi_pkg::resp_t b_resp);
```
zarubaf, line 1455: Maybe it could be helpful if we used a default parameter for `w_prot`. That parameter will most likely be `0` most of the times, right?
```suggestion
    task automatic write(input addr_t w_addr, input prot_t w_prot = 0, input data_t w_data,
                         input strb_t w_strb, output axi_pkg::resp_t b_resp);
```

src/axi_test.sv
```diff
@@ -1435,25 +1451,25 @@ package axi_test;
...
       join
       this.drv.recv_b(b_resp);
-      $display("%0t %s> Recieved write response from ADDR: %h RESP: %h",
+      $display("%0t %s> Received write response from ADDR: %h RESP: %h",
           $time(), this.name, w_addr, b_resp);
     endtask : write
 
     // read data from a specific location
-    task automatic read(input addr_t r_addr,
+    task automatic read(input addr_t r_addr, input prot_t r_prot,
```
zarubaf, line 1468: Same here.

test/tb_axi_lite_mailbox.sv
```diff
@@ -116,56 +116,56 @@ module tb_axi_lite_mailbox;
     // -------------------------------
     $info("Initial test by reading each register");
     $display("%0t MST_0> Read register MBOXW ", $time());
-    lite_axi_master.read(MBOXW, data, resp);
+    lite_axi_master.read(MBOXW, axi_pkg::prot_t'('0), data, resp);
```
zarubaf, line 119: Those could then be replaced by:
```suggestion
    lite_axi_master.read(MBOXW,, data, resp);
```

test/tb_axi_lite_regs.sv
```diff
@@ -0,0 +1,344 @@
...
+  end
+
+  initial begin : proc_generate_direct_load
+    for (int unsigned i = 0; i < RegNumBytes; i++) begin
+      automatic int unsigned j = i;
+      fork
+        begin
+          toggle_load(j);
+        end
+      join_none
```
zarubaf, line 115:
```suggestion
      fork toggle_load(j);
      join_none
```

test/tb_axi_lite_regs.sv
```diff
@@ -0,0 +1,344 @@
...
+    automatic rand_lite_master_t lite_axi_master = new ( master_dv, "Lite Master");
+    automatic axi_data_t      data = '0;
+    automatic axi_pkg::resp_t resp = '0;
+    end_of_sim <= 1'b0;
+    lite_axi_master.reset();
+    @(posedge rst_n);
+    repeat (5) @(posedge clk);
+    lite_axi_master.write(axi_addr_t'(32'h0000_0000), axi_pkg::prot_t'('0),
+        axi_data_t'(64'hDEADBEEFDEADBEEF), axi_strb_t'(8'hFF), resp);
+    lite_axi_master.read(axi_addr_t'(32'h0000_0000), axi_pkg::prot_t'('0), data, resp);
```
zarubaf, line 103: Why are those extra reads/writes necessary? Is that something which isn't covered by the `rand_axi_master`? If yes, wouldn't it make more sense to add this capability to the driver class? If no, why not remove it?

test/tb_axi_lite_regs.sv
```diff
@@ -0,0 +1,344 @@
...
+    end else begin
+      assert (reg_q[byte_i] == save_byte) else
+          $error("AXI write onto read only byte: %0d SAVE: %h EXP: %h ACT %h",
+              byte_i, save_byte, exp_byte, reg_q[byte_i]);
+    end
+  endtask : check_q
+
+  // Some assertions for additional checking.
+  // pragma translate_off
+  `ifndef VERILATOR
```
zarubaf, line 281: You don't need the VERILATOR guard here, we will never elaborate this with verilator :-)

test/tb_axi_lite_regs.sv
WRoenninger: I have these in to quickly check a known register at the beginning of the simulation, to quickly see if something went wrong when tweaking the design.

src/axi_test.sv
WRoenninger: Added that the default is 0.

src/axi_lite_regs.sv
WRoenninger: Added hit to look in the spec.

README.md
accuminium: Good idea, added.

test/tb_axi_lite_regs.sv
accuminium: Then you want to assert the value of `data` and `resp` after the read to do the quick check right? If so, please add the assertions and a short comment on this test case.

WRoenninger: Added comment and assertions.

zarubaf:
LGTM

accuminium:
Great, thanks!

PR #85, opened 2020-04-07
AXI Scoreboard

WRoenninger:
This introduces the class `axi_scoreboard` in `src/axi_test`.
* Add `Monitor` modport in `AXI_BUS_DV`.
* Add `AXI_ASSIGN_MONITOR` macro to assign a `AXI_BUS` interface to a monitor interface.
* Models the accessed memory space of the monitored AXI4+ATOP channel.
  Assumes the memory can only be changed from the AXI channel.
* Handles ambiguous data when reads are performed on memory locations currently written to.

### DIFF
--- a/src/axi_test.sv
+++ b/src/axi_test.sv
@@ -1586,13 +1586,13 @@ package axi_test;
     endtask
   endclass
 
-  /// `axi_scoreboard` class, intended to model a memory space which only gets changed by the
-  /// AXI4+ATOP bus monitored.
-  /// Only capable of modeling the burst type `axi_pkg::BUSRT_INCR`.
-  /// Not capable of atomic operations.
-  /// Internally the memory gets updated over W beats. This scoreboard does NOT support
-  /// reordering of write beat execution, i.e. reordering of B responses on the same
-  /// accesses memory range.
+  /// `axi_scoreboard` models a memory that only gets changed by the monitored AXI4+ATOP bus.
+  ///
+  /// This class is only capable of modeling `INCR` burst type, and cannot handle atomic operations.
+  /// The internal memory representation is updated by W beats.  This class does not support
+  /// reordering of two B responses to one memory location.  That is, if two write transactions on
+  /// the observed bus target the same address, the B responses must be observed in the order of the
+  /// AW beats; otherwise, the internal memory representation of this class gets corrupted.
   ///
   /// Example usage:
   ///   typedef axi_test::axi_scoreboard #(
@@ -1603,9 +1603,9 @@ package axi_test;
   ///   .TT ( TestTime     )
   /// ) axi_scoreboard_t;
   /// axi_scoreboard_t axi_scoreboard = new(monitor_dv);
-  /// initial begin : proc_scoreboard
+  /// initial begin
   ///   axi_scoreboard.enable_all_checks();
-  ///   @(posedge rst_n);
+  ///   wait (rst_n);
   ///   axi_scoreboard.monitor();
   /// end
   class axi_scoreboard #(
--- a/src/axi_test.sv
+++ b/src/axi_test.sv
@@ -1586,6 +1586,405 @@ package axi_test;
     endtask
   endclass
 
+  /// `axi_scoreboard` class, intended to model a memory space which only gets changed by the
+  /// AXI4+ATOP bus monitored.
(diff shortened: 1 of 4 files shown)


src/axi_test.sv
```diff
@@ -1569,6 +1569,400 @@ package axi_test;
...
+    /// Handle update of the golden model with W beats.
+    protected task automatic handle_write();
+      axi_addr_t beat_addresses [];
+      axi_addr_t bus_address;
+      ax_beat_t  aw_beat;
+      w_beat_t   w_beat;
+      byte_t     write_data;
+      forever begin
+        wait (this.aw_sample.size() > 0);
+        aw_beat        = this.aw_sample.pop_front();
```
accuminium, line 1678: Please `assert` here that `aw_beat` is `INCR` and not an ATOP and `$error` (or `$fatal` since the model breaks?) else.

I later saw those assertions below, but I think it's better to move them here, since *this* is the code that has restrictions on the `aw_beat` -- not the monitor below.

src/axi_test.sv
```diff
@@ -1569,6 +1569,400 @@ package axi_test;
...
+    protected task automatic handle_read(input axi_id_t id);
+      ax_beat_t  ar_beat;
+      r_beat_t   r_beat;
+      axi_addr_t bus_address, beat_address, idx_data;
+      byte_t     act_data;
+      byte_t     exp_data[$];
+      byte_t     tst_data[$];
+      forever begin
+        wait (this.ar_sample[id].size() > 0);
+        ar_beat = this.ar_sample[id].pop_front();
```
accuminium, line 1743: Please `assert` here that `ar_beat` is `INCR` (and act as you decide above if it is not).

src/axi_test.sv
```diff
@@ -1569,6 +1569,400 @@ package axi_test;
...
+          aw_beat.ax_prot   = this.axi.aw_prot;
+          aw_beat.ax_qos    = this.axi.aw_qos;
+          aw_beat.ax_region = this.axi.aw_region;
+          aw_beat.ax_atop   = this.axi.aw_atop;
+          aw_beat.ax_user   = this.axi.aw_user;
+          assert (aw_beat.ax_burst == axi_pkg::BURST_INCR ||
+                  aw_beat.ax_len   == '0                     ) else
+              $warning("Not supported AW burst: BURST: %0h.", aw_beat.ax_burst);
+          assert (aw_beat.ax_atop == '0) else
+              $warning("Atomic transfers not supported: ATOP: %0h.", aw_beat.ax_atop);
```
accuminium, line 1800: I suggest to move those two `assert`s as discussed above. Also `$warning` seems too weak? Is the state of the scoreboard consistent after one of those conditions is violated?

src/axi_test.sv
```diff
@@ -1569,6 +1569,400 @@ package axi_test;
...
+          ar_beat.ax_lock   = this.axi.ar_lock;
+          ar_beat.ax_cache  = this.axi.ar_cache;
+          ar_beat.ax_prot   = this.axi.ar_prot;
+          ar_beat.ax_qos    = this.axi.ar_qos;
+          ar_beat.ax_region = this.axi.ar_region;
+          ar_beat.ax_atop   = 6'bxxxxxx;
+          ar_beat.ax_user   = this.axi.ar_user;
+          assert (ar_beat.ax_burst == axi_pkg::BURST_INCR ||
+                  ar_beat.ax_len   == '0                     ) else
+              $warning("Not supported AR burst: BURST: %0h.", ar_beat.ax_burst);
```
accuminium, line 1861: See above.

src/axi_test.sv
WRoenninger: It will handle all transactions as `BURST_INCR`.
However with the changes from #94 it would be possible to handle all burst types in the scoreboard.

accuminium:
I rebased on `master`; let's wait for CI to pass: [CI status](https://iis-git.ee.ethz.ch/akurth/axi/commits/axi_scoreboard).

WRoenninger:
CI will probably fail, as the scoreboard uses `axi_pkg::beat_addr` which was changed in #94.

accuminium:
> CI will probably fail, as the scoreboard uses `axi_pkg::beat_addr` which was changed in #94.

Indeed it did. I fixed that and it is running again.

PR #89, opened 2020-04-08
axi_lite_to_apb: Fix strobe width assertion message

### LINKED ISSUE #87: Wrong assertion string
https://github.com/pulp-platform/axi/blob/8f0de0a90be33c3e75628a343df06ca7dbd293e4/src/axi_lite_to_apb.sv#L318

This should say something about the strobe width not being equal, not the address width.

accuminium:
Fixes #87.


accuminium:
Splendid, thanks!

PR #93, opened 2020-04-09
AXI Serializer

WRoenninger:
* Adds `axi_serializer` from #91.
* Adds ATOP support to the module `axi_serializer`.
* Functional and Synthesis testbench for `axi_serializer`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## Unreleased
 
 ### Added
+- Add `axi_serializer`, to serialize multiple transactions with different IDs to the same ID.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## Unreleased
 
 ### Added
-- Add `axi_serializer`, to serialize multiple transactions with different IDs to the same ID.
+- `axi_serializer`: serialize transactions with different IDs to the same ID.
 
 ### Changed
 
--- a/src/axi_serializer.sv
+++ b/src/axi_serializer.sv
@@ -184,19 +184,19 @@ module axi_serializer #(
   default disable iff (~rst_ni);
   aw_lost : assert property( @(posedge clk_i)
       (slv_req_i.aw_valid & slv_resp_o.aw_ready |-> mst_req_o.aw_valid & mst_resp_i.aw_ready))
-    else $error ("AW beat lost.");
+    else $error("AW beat lost.");
   w_lost  : assert property( @(posedge clk_i)
       (slv_req_i.w_valid & slv_resp_o.w_ready |-> mst_req_o.w_valid & mst_resp_i.w_ready))
-    else $error ("W beat lost.");
+    else $error("W beat lost.");
   b_lost  : assert property( @(posedge clk_i)
       (mst_resp_i.b_valid & mst_req_o.b_ready |-> slv_resp_o.b_valid & slv_req_i.b_ready))
-    else $error ("B beat lost.");
+    else $error("B beat lost.");
   ar_lost : assert property( @(posedge clk_i)
       (slv_req_i.ar_valid & slv_resp_o.ar_ready |-> mst_req_o.ar_valid & mst_resp_i.ar_ready))
-    else $error ("AR beat lost.");
+    else $error("AR beat lost.");
   r_lost :  assert property( @(posedge clk_i)
       (mst_resp_i.r_valid & mst_req_o.r_ready |-> slv_resp_o.r_valid & slv_req_i.r_ready))
-    else $error ("R beat lost.");
+    else $error("R beat lost.");
 `endif
 // pragma translate_on
 endmodule
--- a/src/axi_serializer.sv
+++ b/src/axi_serializer.sv
@@ -46,12 +46,12 @@ module axi_serializer #(
     AtopExecute = 2'b10
   } state_e;
 
-  logic                      rd_fifo_full, rd_fifo_empty, rd_fifo_push,
-                             wr_fifo_full, wr_fifo_empty, wr_fifo_push;
-  id_t                       b_id,
-                             r_id,         ar_id;
-  state_e                    state_q,      state_d;
-  logic                      change_state;
+  logic   rd_fifo_full, rd_fifo_empty, rd_fifo_push, rd_fifo_pop,
+          wr_fifo_full, wr_fifo_empty, wr_fifo_push, wr_fifo_pop;
+  id_t    b_id,
+          r_id,         ar_id;
+  state_e state_q,      state_d;
+  logic   change_state;
 
   always_comb begin
     // Default assignments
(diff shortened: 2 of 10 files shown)


src/axi_serializer.sv
```diff
@@ -9,121 +9,238 @@
...
 
-    // Gate R handshake with valid output of Read FIFO.
-    slv_resp_o.r_valid = mst_resp_i.r_valid & rd_fifo_valid;
-    mst_req_o.r_ready = slv_req_i.r_ready & rd_fifo_valid;
+    // Gate R handshake with empty flag output of Read FIFO.
+    slv_resp_o.r_valid = mst_resp_i.r_valid & ~rd_fifo_empty;
+    mst_req_o.r_ready  = slv_req_i.r_ready  & ~rd_fifo_empty;
   end
+  // State change enable
+  assign change_state = state_d != state_q;
```
accuminium, line 140: What is this for? Why not simply use a `FFARN`?

src/axi_serializer.sv
```diff
@@ -9,121 +9,238 @@
...
+`ifndef VERILATOR
+  initial begin: p_assertions
+    assert (AxiIdWidth    >= 1) else $fatal(1, "AXI ID   width must be at least 1!");
+    assert (MaxReadTxns   >= 1)
+      else $fatal(1, "Maximum number of read transactions must be >= 1!");
+    assert (MaxWriteTxns  >= 1)
+      else $fatal(1, "Maximum number of write transactions must be >= 1!");
+  end
+  aw_lost : assert property(
+    @(posedge clk_i) disable iff (~rst_ni)
```
accuminium, line 190: Please simplify with `default disable`.

src/axi_serializer.sv
```diff
@@ -9,121 +9,238 @@
...
+    .pop_i      ( slv_resp_o.b_valid & slv_req_i.b_ready    ),
+    .usage_o    ( wr_usage                                  )
   );
 
+  `FFLARN(state_q, state_d, change_state, AtopIdle, clk_i, rst_ni)
+
+// pragma translate_off
+`ifndef VERILATOR
+  initial begin: p_assertions
+    assert (AxiIdWidth    >= 1) else $fatal(1, "AXI ID   width must be at least 1!");
```
accuminium, line 183:
```suggestion
    assert (AxiIdWidth    >= 1) else $fatal(1, "AXI ID width must be at least 1!");
```

src/axi_serializer.sv
```diff
@@ -9,121 +9,238 @@
...
+    assert (MaxReadTxns   >= 1)
+      else $fatal(1, "Maximum number of read transactions must be >= 1!");
+    assert (MaxWriteTxns  >= 1)
+      else $fatal(1, "Maximum number of write transactions must be >= 1!");
+  end
+  aw_lost : assert property(
+    @(posedge clk_i) disable iff (~rst_ni)
+        (slv_req_i.aw_valid & slv_resp_o.aw_ready |-> mst_req_o.aw_valid & mst_resp_i.aw_ready))
+    else $fatal (1, "AW beat lost.");
```
accuminium, line 192:
```suggestion
    else $error (1, "AW beat lost.");
```
Please change these "lost" assertions to `error`s.

test/tb_axi_serializer.sv
```diff
@@ -0,0 +1,189 @@
...
+  localparam time CyclTime = 10ns;
+  localparam time ApplTime =  2ns;
+  localparam time TestTime =  8ns;
+  // AXI configuration
+  localparam int unsigned AxiIdWidth   =  4;
+  localparam int unsigned AxiAddrWidth =  32;    // Axi Address Width
+  localparam int unsigned AxiDataWidth =  64;    // Axi Data Width
+  localparam int unsigned AxiUserWidth =  5;
+  // Sim print config, how many transactions
+  localparam int unsigned PrintTnx = 500;
```
accuminium, line 36:
```suggestion
  localparam int unsigned PrintTxn = 500;
```

test/tb_axi_serializer.sv
```diff
@@ -0,0 +1,189 @@
...
+    $stop();
+  end
+
+  initial begin : proc_axi_slave
+    automatic rand_axi_slave_t  rand_axi_slave  = new(slave_dv);
+    rand_axi_slave.reset();
+    @(posedge rst_n);
+    rand_axi_slave.run();
+  end
+
```
accuminium, line 148: Please add some model checking logic and assertions:

- Assert that downstream Ax equal upstream Ax except with ID 0.
- Assert that downstream W equal upstream W.
- Assert that upstream B and R equal downstream B and R with reflected ID.

src/axi_serializer.sv
```diff
@@ -9,121 +9,238 @@
...
+    // Default, connect the channels
+    mst_req_o  = slv_req_i;
+    slv_resp_o = mst_resp_i;
 
     // Serialize transactions -> tie downstream IDs to zero.
     mst_req_o.aw.id = '0;
     mst_req_o.ar.id = '0;
 
     // Reflect upstream ID in response.
+    aw_id           = slv_req_i.aw.id;
```
accuminium, line 75: `aw_id` is always `slv_req_i.aw.id`, correct? If so, please change to `assign` outside `always_comb`.

src/axi_serializer.sv
```diff
@@ -9,121 +9,238 @@
...
+          if (mst_resp_i.aw_ready) begin
+            state_d = AtopExecute;
+          end
+        end
+      end
+      AtopExecute:    begin
+        if (wr_fifo_empty && rd_fifo_empty) begin
+          state_d = AtopIdle;
+        end
+      end
```
accuminium, line 127: Please fuse this case with the `AtopIdle` case (adding conditions to distinguish the two cases where necessary) to allow processing the following transaction already in `AtopExecute` without duplicating code.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## Unreleased
 
 ### Added
+- Add `axi_serializer`, to serialize multiple transactions with different IDs to the same ID.
```
accuminium, line 11:
```suggestion
- `axi_serializer`: serialize transactions with different IDs to the same ID.
```

src/axi_serializer.sv
```diff
@@ -9,121 +9,238 @@
...
+  } state_e;
 
-  logic rd_fifo_ready,  rd_fifo_valid,
-        wr_fifo_ready,  wr_fifo_valid;
-  id_t  b_id,
-        r_id;
+  logic                      rd_fifo_full, rd_fifo_empty, rd_fifo_push,
+                             wr_fifo_full, wr_fifo_empty, wr_fifo_push;
+  logic   [RdUsageWidth-1:0] rd_usage;
+  logic   [WrUsageWidth-1:0] wr_usage;
```
accuminium, line 54: Please remove those two unused signals (also from the wave file) and leave the corresponding FIFO outputs open.

src/axi_serializer.sv
```diff
@@ -9,121 +9,238 @@
...
+    mst_req_o.ar_valid  = 1'b0;
+    slv_resp_o.ar_ready = 1'b0;
+    mst_req_o.aw_valid  = 1'b0;
+    slv_resp_o.aw_ready = 1'b0;
 
-    // Gate B handshake with valid output of Write FIFO.
-    slv_resp_o.b_valid = mst_resp_i.b_valid & wr_fifo_valid;
-    mst_req_o.b_ready = slv_req_i.b_ready & wr_fifo_valid;
+    unique case (state_q)
+      AtopIdle:    begin
```
accuminium, line 87:
```suggestion
      AtopIdle: begin
```
No need to indent `begin`, single space suffices (also for the other cases).

src/axi_serializer.sv
WRoenninger: As the state change happens rarely in normal operation, I think it would be beneficial to explicit enable the FF. So that the synthesis toll will always refer a clock gate for the FSM.

test/tb_axi_serializer.sv
WRoenninger: Added checker in TB

src/axi_serializer.sv
WRoenninger: However this would not lead to a reduction in state holding elements? I would still need a FF to know if there is an active ATOP downstream.

src/axi_serializer.sv
accuminium: Explicitly clock-gating two FFs in the FSM strikes me as odd in terms of optimization priorities. Depending on the implementation of the FIFOs, there are many more FFs in there, so we would focus on that if we want to optimize clock gating of this module.

Additionally, I think any synthesizer with reasonable automatic clock gating should be able to derive the `state_d != state_q` condition.

@zarubaf: What are your experiences and opinion on this?

src/axi_serializer.sv
accuminium: No, it does not reduce state holding elements, but it allows us to share the code for accepting transactions and going into the `AtopDrain` state between the `AtopIdle` and the `AtopExecute` state.

Something like this:
```sv
AtopIdle, AtopExecute: begin
  // some common code
  if (state_q == AtopIdle) begin
    // code specific to AtopIdle
  end
  if (state_q == AtopExecute) begin
    // code specific to AtopExecute
  end
  // ...
end
```

src/axi_serializer.sv
zarubaf: I completely agree with @accuminium. The synthesis tool can also infer state changes from `always_comb` blocks, and if necessary, derive clock gates. It is anyway not given, that a clock gate will be inferred by using the load-enable flipflop. For example, for these two flipflops it might not even pay-off to infer a clock gate (another cell which consumes static power, etc.).

Long story short, my personal opinion: 
- Code readability has priority, which IMHO is inhibited in this case.
- Load-enables for larger structures and flipflops which occasionally latch data.

accuminium:
CI for `axi_serializer` already looks good: [CI status](https://iis-git.ee.ethz.ch/akurth/axi/commits/axi_serializer)

zarubaf:
I have one additional request regarding this module (and a point for discussion). The module currently pretty much expects atomics on the bus. We break compatibility with AXI buses which don't have atomics. While I see that most of our infrastructure is tuned for the atop, this module would make a good candidate to make the `atop` feature optional, using a parameter. If we don't want to solve this here, we can also move this to a separate issue.

accuminium:
How do we break compatibility with AXI buses that do not have atomics?

zarubaf:
By access to the `atop` signal:
https://github.com/WRoenninger/axi/blob/782f8be2ff45dfe84e62c187b4b5bc71d3d469ba/src/axi_serializer.sv#L100

This regards AXI Buses which don't have this signal (AXI-ACE for example).

accuminium:
Granted, we require the `atop` signal to be present in the request struct and tied off to zero if unused. But this is in-line with our top-level bus requirements: https://github.com/pulp-platform/axi#atomic-operations.

accuminium:
I will rebase to clean up and give it another CI run.

PR #33, opened 2020-01-11
WIP: AXI4 Last Level Cache

WRoenninger:
Added AXI4 Last Level Cache
Features: 
- Write back, Read and Write allocate
- Parameterizable in set-associativity, line length and number of blocks
- Adaptive scratch pad memory mapping of individual sets
- Multiple open AXI transactions
- Tag memory initialization BIST
- Flush by user and SPM reconfiguration
- Random replacement
Added Documentation

This PR directly depends on changes from #116 and #115.

PR #95, opened 2020-04-16
Bender: Bump common_cells to 1.18.0

WRoenninger:
Some new streaming modules from `common_cells` will be needed for #33.


accuminium:
Can you please bump dependencies in the PR that needs them? Otherwise tracking which change introduces which dependency is painful.

PR #98, opened 2020-04-24
WIP: AXI4+ATOP Translation Lookaside Buffer (TLB)

andreaskurth:
Here is a first draft of an AXI TLB. I would highly appreciate initial architectural feedback.

The concept is similar to that of the RAB (no need to reinvent the wheel), but in a fully stream-based, modular implementation (using the `common_cells` stream infrastructure) and leveraging core `axi` modules such as `axi_demux` to handle the protocol.

## TODOs
- [ ] Integrate `axi_lite_regs` from #79.
- [ ] Testbench
- [ ] Documentation

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -48,9 +48,11 @@ sources:
   - src/axi_err_slv.sv
   - src/axi_dw_converter.sv
   - src/axi_multicut.sv
+  - src/axi_tlb_l1.sv
   - src/axi_to_axi_lite.sv
   # Level 4
   - src/axi_lite_xbar.sv
+  - src/axi_tlb.sv
   - src/axi_xbar.sv
 
   - target: synth_test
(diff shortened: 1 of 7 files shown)


accuminium:
Here is a first draft of an AXI TLB. I would highly appreciate initial architectural feedback.

The concept is similar to that of the RAB (no need to reinvent the wheel), but in a fully stream-based, modular implementation (using the `common_cells` stream infrastructure) and leveraging core `axi` modules such as `axi_demux` to handle the protocol.

zarubaf:
I don't have any objections to the hardware architecture. I'd probably designed it very similarly. The stream-based description approach makes the RTL very, very readable. Nice work.

micprog:
The AXI Translation Lookaside Buffer (TLB) has been moved to a dedicated repository on pulp-platform: [`axi_tlb`](https://github.com/pulp-platform/axi_tlb)

PR #99, opened 2020-04-24
axi_modify_address: Various fixes and improvements

accuminium:
This fixes a variety of problems and shortcomings of `axi_modify_address` and its interface variant; details are in the Changelog.

## Remaining TODOs
- [x] Add synthesis bench.
- [x] Add simple testbench.

### DIFF
--- a/src/axi_modify_address.sv
+++ b/src/axi_modify_address.sv
@@ -21,12 +21,12 @@ module axi_modify_address #(
   /// Request type of the master port
   parameter type  mst_req_t = logic,
   /// Response type of slave and master port
-  parameter type     resp_t = logic
+  parameter type axi_resp_t = logic
 ) (
   /// Slave port request
   input  slv_req_t  slv_req_i,
   /// Slave port response
-  output resp_t     slv_resp_o,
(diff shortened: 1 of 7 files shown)


src/axi_modify_address.sv
```diff
@@ -12,29 +12,30 @@
...
-  parameter type  mst_req_t = logic, // request type master port
-  parameter type mst_resp_t = logic  // response type master port
+  /// Request type of the slave port
+  parameter type  slv_req_t = logic,
+  /// Address type of the master port
+  parameter type mst_addr_t = logic,
+  /// Request type of the master port
+  parameter type  mst_req_t = logic,
+  /// Response type of slave and master port
+  parameter type     resp_t = logic
```
WRoenninger, line 24: Would it be better to have the name changed to avoid name collision with `axi_pkg::resp_t` if someone integrating this module decides to have a global `import axi_pkg::*`?

```suggestion
  parameter type     axi_resp_t = logic
```

accuminium: That's a very good point, thanks!

PR #100, opened 2020-04-27
Add support for FIXED transactions in the DWCs

suehtamacv:
This PR adds support for `FIXED` transactions in the DWCs. 

It is trivial to upsize `FIXED` transactions, and the upsized AXI transaction is identical to the original. 
Downsizing them, however, is more complex:
-  If the incoming `FIXED` transactions have an `axsize` not wider than the outcoming port, the DWC can forward the transaction without any changes.
- Otherwise, each data beat of the incoming `FIXED`  transaction must be transformed into an `INCR` burst.

The downsizer currently does not support `FIXED` bursts. It would be possible to support them by converting them into a sequence of `INCR` bursts. But I wonder whether this is a common enough use case to require the support of the downsizer. If such bursts are needed, an alternative would be to use an `axi_burst_splitter` in front of the downsizer.

### DIFF
--- a/src/axi_dw_downsizer.sv
+++ b/src/axi_dw_downsizer.sv
@@ -737,7 +737,7 @@ module axi_dw_downsizer #(
             w_req_d.aw_throw_error = 1'b1         ;
 
             // ... but might if this is a narrow single-beat transaction
-            if (slv_req_i.aw.size <= AxiSlvPortMaxSize && slv_req_i.aw.len == '0)
+            if (slv_req_i.aw.size <= AxiMstPortMaxSize && slv_req_i.aw.len == '0)
               w_req_d.aw_throw_error = 1'b0;
           end
 
@@ -747,7 +747,7 @@ module axi_dw_downsizer #(
             w_req_d.aw_throw_error = 1'b1         ;
 
             // ... but might if this is a narrow single-beat transaction
-            if (slv_req_i.aw.size <= AxiSlvPortMaxSize && slv_req_i.aw.len == '0)
+            if (slv_req_i.aw.size <= AxiMstPortMaxSize && slv_req_i.aw.len == '0)
               w_req_d.aw_throw_error = 1'b0;
           end
         endcase
--- a/src/axi_dw_downsizer.sv
+++ b/src/axi_dw_downsizer.sv
@@ -15,8 +15,9 @@
 // Connects a wide master to a narrower slave.
 
 // NOTE: The downsizer does not support WRAP bursts, and will answer with SLVERR
-// upon receiving a burst of such type. The downsizer also does not support
-
+// upon receiving a burst of such type.  The downsizer does support FIXED
+// bursts, but only if they consist of a single beat; it will answer with SLVERR
+// on multi-beat FIXED bursts.
 module axi_dw_downsizer #(
     parameter int unsigned AxiMaxReads         = 1    , // Number of outstanding reads
     parameter int unsigned AxiSlvPortDataWidth = 8    , // Data width of the slv port
--- a/src/axi_dw_downsizer.sv
+++ b/src/axi_dw_downsizer.sv
@@ -14,8 +14,8 @@
 // Data width downsize conversion.
 // Connects a wide master to a narrower slave.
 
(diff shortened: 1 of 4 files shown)


accuminium:
Thanks for these changes, @suehtamacv. Let's split this in three parts:
1) Fixes to the current DWC features and improvements of the DWC verification infrastructure. I think these are ready to be merged directly and necessary for adding model checking to the TBs.
    - :bug: axi_dw_downsizer: Let narrow AXI transitions pass through
    - axi_dw_converter: Clean up testbenches
2) Helper functions in `axi_pkg`.  These are fairly uncontroversial as well, but I would like to have brief input from other devs on these.
    - axi_pkg: Add modifiable and bufferable helper functions
3) Support for *fixed* bursts in the DWC. These are more intricate and I would like to have proper TBs before doing major changes to DWC code.
    - axi_dw_upsizer: Implement upsizing of FIXED burst transactions
    - axi_dw_downsizer: Implement downsizing of FIXED single-beat burst transactions

accuminium:
> This PR adds support for `FIXED` transactions in the DWCs.
> 
> It is trivial to upsize `FIXED` transactions, and the upsized AXI transaction is identical to the original.
> Downsizing them, however, is more complex:
> 
>     * If the incoming `FIXED` transactions have an `axsize` not wider than the outcoming port, the DWC can forward the transaction without any changes.
> 
>     * Otherwise, each data beat of the incoming `FIXED`  transaction must be transformed into an `INCR` burst.
> 
> 
> The downsizer currently does not support `FIXED` bursts. It would be possible to support them by converting them into a sequence of `INCR` bursts. But I wonder whether this is a common enough use case to require the support of the downsizer. If such bursts are needed, an alternative would be to use an `axi_burst_splitter` in front of the downsizer.

So you are suggesting
1) Add support for `FIXED` transactions to the upsizer, since that is trivial.
2) Add support for `FIXED` transactions with `axsize <= AxiMstPortDataWidth` to the downsizer, since that is trivial as well.
3) Keep rejecting `FIXED` transactions with `axsize > AxiMstPortDataWidth` in the downsizer, and document this limitation mentioning the option of an `axi_burst_splitter` before the downsizer if this case must be handled.

Right?

suehtamacv:
No, my point was to reject `FIXED` _bursts_ (with `axlen > 0`). Accordingly to [Xilinx](https://www.xilinx.com/support/documentation/ip_documentation/axi_interconnect/v2_1/pg059-axi-interconnect.pdf), page 82, a `FIXED` burst with downsize ratio greater than one should be converted to `axlen + 1` `INCR` bursts. Again, if this is needed, I can add this feature. I actually modeled this behavior at the golden model of the downsizer, #102.

Back to the downsizer, I would like to bring a point from the AXI specification (IHI0022H, page A4-65) about non-modifiable transactions:

>  There are circumstances where it is not possible to meet the requirements of Non-modifiable transactions. For example, when downsizing to a bus width narrower than that required by the transaction size, AxSIZE, the transaction must be modified. A component that performs such an operation can optionally include an IMPLEMENTATION DEFINED mechanism to indicate that a modification has occurred. This mechanism can assist with software debug.

The way I understand, it is therefore technically possible to downsize _non-modifiable_ transactions that are wider than the narrow slave bus. Currently, the downsizer rejects such requests, either `INCR` or `FIXED`. Should I ignore the non-modifiable bit of transactions arriving at the downsizer?

accuminium:
I am adding a Changelog entry. Please let me know if you agree with it. Then this can be merged.

src/axi_dw_downsizer.sv
```diff
@@ -14,8 +14,8 @@
 // Data width downsize conversion.
 // Connects a wide master to a narrower slave.
 
-// NOTE: The downsizer does not support WRAP and FIXED bursts, and
-// will answer with SLVERR upon receiving a burst of such types.
+// NOTE: The downsizer does not support WRAP bursts, and will answer with SLVERR
+// upon receiving a burst of such type. The downsizer also does not support
```
suehtamacv, line 18: Can you fix this extraneous comment?

PR #102, opened 2020-04-29
Add golden models for the DW Converters

suehtamacv:
This adds two monitors in the `tb_axi_dw_pkg.sv` file, against which the DW converters are compared on the TB.

### DIFF
--- a/src/axi_dw_downsizer.sv
+++ b/src/axi_dw_downsizer.sv
@@ -787,78 +787,80 @@ module axi_dw_downsizer #(
       w_req_d.aw_throw_error = 1'b0              ;
       w_req_d.burst_resp     = axi_pkg::RESP_OKAY;
 
-      if (slv_req_i.aw_valid && slv_req_i.aw.atop[5]) begin // ATOP with an R response
-        inject_aw_into_ar_req = 1'b1                 ;
-        slv_resp_o.aw_ready   = inject_aw_into_ar_gnt;
-      end else begin // Regular AW
-        slv_resp_o.aw_ready = 1'b1;
-      end
+      if (!forward_b_beat_full) begin
+        if (slv_req_i.aw_valid && slv_req_i.aw.atop[5]) begin // ATOP with an R response
+          inject_aw_into_ar_req = 1'b1                 ;
+          slv_resp_o.aw_ready   = inject_aw_into_ar_gnt;
+        end else begin // Regular AW
+          slv_resp_o.aw_ready = 1'b1;
+        end
 
-      // New write request
-      if (slv_req_i.aw_valid && slv_resp_o.aw_ready && !forward_b_beat_full) begin
-        // Default state
-        w_state_d = W_PASSTHROUGH;
+        // New write request
+        if (slv_req_i.aw_valid && slv_resp_o.aw_ready) begin
+          // Default state
+          w_state_d = W_PASSTHROUGH;
 
-        // Save beat
-        w_req_d.aw            = slv_req_i.aw      ;
-        w_req_d.aw_valid      = 1'b1              ;
-        w_req_d.burst_len     = slv_req_i.aw.len  ;
-        w_req_d.orig_aw_len   = slv_req_i.aw.len  ;
-        w_req_d.orig_aw_size  = slv_req_i.aw.size ;
-        w_req_d.orig_aw_burst = slv_req_i.aw.burst;
+          // Save beat
+          w_req_d.aw            = slv_req_i.aw      ;
+          w_req_d.aw_valid      = 1'b1              ;
+          w_req_d.burst_len     = slv_req_i.aw.len  ;
+          w_req_d.orig_aw_len   = slv_req_i.aw.len  ;
+          w_req_d.orig_aw_size  = slv_req_i.aw.size ;
+          w_req_d.orig_aw_burst = slv_req_i.aw.burst;
 
-        case (slv_req_i.aw.burst)
-          axi_pkg::BURST_INCR: begin
-            // Evaluate downsize ratio
-            automatic addr_t size_mask  = (1 << slv_req_i.aw.size) - 1                                              ;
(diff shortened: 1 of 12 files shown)


test/tb_axi_dw_pkg.sv
```diff
@@ -0,0 +1,852 @@
...
+        end
+
+        // Populate the expected B queue
+        exp_b = '{axi_id: master_axi.aw_id, last: 1'b1};
+        this.exp_b_queue.push(master_axi.aw_id, exp_b);
+        incr_expected_tests(1)                        ;
+        $display("        Expect B response.")        ;
+
+        // Inject expected R beats on this id, if it is an atop
+        if(master_axi.aw_atop[5]) begin
```
WRoenninger, line 468: We have a constant in `axi_pkg` which defines the bit to listen to for ATOP read responses, makes it more readable
```suggestion
        if(master_axi.aw_atop[axi_pkg::ATOP_R_RESP]) begin
```

test/tb_axi_dw_upsizer.sv
```diff
@@ -15,159 +15,169 @@
...
+    // AXI Parameters
+    parameter int unsigned AxiAddrWidth        = 64  ,
+    parameter int unsigned AxiIdWidth          = 4   ,
+    parameter int unsigned AxiSlvPortDataWidth = 32  ,
+    parameter int unsigned AxiMstPortDataWidth = 64  ,
+    parameter int unsigned AxiUserWidth        = 8   ,
+    // TB Parameters
+    parameter time CyclTime                    = 10ns,
+    parameter time ApplTime                    = 0ns ,
+    parameter time TestTime                    = 5ns
```
WRoenninger, line 28: Same as in `tb_axi_dw_sownsizer`:
```suggestion
    parameter time ApplTime                    = 2ns ,
    parameter time TestTime                    = 8ns
```

test/tb_axi_dw_upsizer.sv
```diff
@@ -15,159 +15,169 @@
...
+    master_drv.add_memory_region({AxiAddrWidth{1'b0}}, {AxiAddrWidth{1'b1}}, axi_pkg::WTHRU_NOALLOCATE);
+
+    fork
+      // Act as a sink
+      slave_drv.run()         ;
+      master_drv.run(200, 200);
+    join_any
+
+    // Done
+    #(10*CyclTime);
```
WRoenninger, line 153: Same as in `tb_axi_downsizer`:
```suggestion
    repeat (10) @(posedge clk);
```

test/tb_axi_dw_pkg.sv
accuminium: Would this be worth a function, like recently defined for `modifiable` and `bufferable`?

test/tb_axi_dw_pkg.sv
```diff
@@ -0,0 +1,852 @@
...
+    // the corresponding ID queue with the number of R beats indicated on the `ar_len` field.
+    // Emphasis on the last flag.
+    task automatic monitor_mst_ar ();
+      exp_ax_t exp_slv_ar;
+      exp_t    exp_mst_r;
+
+      if (master_axi.ar_valid && master_axi.ar_ready) begin
+        case (master_axi.ar_burst)
+          axi_pkg::BURST_INCR: begin
+            automatic int unsigned downsize_ratio = ((1'b1 << master_axi.ar_size) + AxiMstPortStrbWidth - 1) / AxiMstPortStrbWidth ;
```
accuminium, line 735: Please refactor the code shared between reads and writes (e.g., to calculate `downsize_ratio`, `size_mask`, etc.) into functions to minimize redundancy.

test/tb_axi_dw_pkg.sv
suehtamacv: Fixed. It is a common enough use case, but unlike `bufferable` and `modifiable`, which had a bit mask that had to be AND-ed with the value of `axcache`, `ATOP_R_RESP` indicates the bit responsible for holding the important information. I guess it would make sense to align both under the same convention.

suehtamacv:
The latest commits improve coverage by checking the data fields of the R and W channels. Due to race conditions between channels, the monitors don't operate on the channels directly, but rather on enqueued beats. A new task, `store_channels()`, enqueues any beats that it sees on the AR, AW, R and W channels in the respective queue.

src/axi_dw_downsizer.sv
```diff
@@ -66,6 +68,10 @@ module axi_dw_downsizer #(
   localparam SlvPortByteMask = AxiSlvPortStrbWidth - 1;
   localparam MstPortByteMask = AxiMstPortStrbWidth - 1;
 
+  // Byte-grouped data words
+  typedef logic [AxiMstPortDataWidth-1:0][7:0] mst_data_t;
+  typedef logic [AxiSlvPortDataWidth-1:0][7:0] slv_data_t;
```
accuminium, line 73: @suehtamacv Should the outer dimension not be `...StrbWidth` instead of `...DataWidth`?

src/axi_dw_upsizer.sv
```diff
@@ -63,6 +65,10 @@ module axi_dw_upsizer #(
   localparam AxiSlvPortMaxSize = $clog2(AxiSlvPortStrbWidth);
   localparam AxiMstPortMaxSize = $clog2(AxiMstPortStrbWidth);
 
+  // Byte-grouped data words
+  typedef logic [AxiMstPortDataWidth-1:0][7:0] mst_data_t;
+  typedef logic [AxiSlvPortDataWidth-1:0][7:0] slv_data_t;
```
accuminium, line 70: Also here

src/axi_dw_downsizer.sv
suehtamacv: Indeed, let me fix this in a second

suehtamacv: Fixed

accuminium (approved):

src/axi_dw_downsizer.sv
accuminium: Thanks!

PR #106, opened 2020-05-06
rand_axi_master: Respect burst type restrictions in ATOPs

### LINKED ISSUE #104: random_axi_master issues WRAP transactions regardless of AXI_BURST_WRAP
When configured with `AXI_ATOPS = 1'b1`, the `random_axi_master` might issue `WRAP` transactions regardless of the value of `AXI_BURST_WRAP`. 

https://github.com/pulp-platform/axi/blob/b98dd67009224076c26b3dc06ceda8b8dcd52b1a/src/axi_test.sv#L880-L889

This creates an issue with modules that do not support `WRAP` transactions and use `random_axi_master` in their TBs. (The same thing can be said about `BURST_INCR`, although this is a more common case).

accuminium:
This fixes #104.

CI status

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


accuminium:
Thanks!

PR #109, opened 2020-05-11
Fusesoc

olofk:
Fix a couple of warnings and add a FuseSoC-compatible AXI interconnect generator


axi_intercon_gen.py
```diff
@@ -0,0 +1,552 @@
...
+                if ch in ['ar', 'aw'] and name == 'id' and m.idw < max_idw:
+                    src = "{"+ str(max_idw-m.idw)+"'d0,"+src+"}"
+                if m.read_only and (ch in ['aw' , 'w', 'b']):
+                    if ch == 'aw' and name == 'id':
+                        _w = max_idw
+                    else:
+                        _w = max(1,width)
+                    src = "{}'d0".format(_w)
+                raw += "   assign masters_req[{}].{} = {};\n".format(i, _name, src)
+#        raw += "   assign connectivity_map[{}] = {}'b{};\n".format(i, len(slaves), '1'*len(slaves))
```
accuminium, line 186: Can code that is commented out be removed?

axi_intercon_gen.py
```diff
@@ -0,0 +1,552 @@
...
+    signals = [
+        ("awid"    , False, id_width),
+        ("awaddr"  , False, w.addr  ),
+        ("awlen"   , False, 8  ),
+        ("awsize"  , False, 3  ),
+        ("awburst" , False, 2 ),
+        ("awlock"  , False, 0 ),
+        ("awcache" , False, 4 ),
+        ("awprot"  , False, 3 ),
+        ("awregion", False, 4),
```
accuminium, line 82: We have an additional `awatop` port for AXI5 ATOPs, does that need to be added to this list?

axi_intercon_gen.py
```diff
@@ -0,0 +1,552 @@
+#!/usr/bin/env python
```
accuminium, line 1: Can we move the two Python scripts to the `scripts` directory? That's where we keep tool-specific scripts.

axi_intercon_gen.py
olofk: Ouch. I see now that the file is a bit messy. Will sleep before pushing next time :)

axi_intercon_gen.py
olofk: Sure thing

axi_intercon_gen.py
olofk: I ignored awatop, but you're right. It should be added. Since it's probably not commonly used, I would prefer to have it optional. I can add a config switch to enable it

olofk:
Thanks for the review. Will clean up commented code, add awatop support and move to the scripts dir. Sending an updated patch after that

accuminium:
> Thanks for the review. Will clean up commented code, add awatop support and move to the scripts dir. Sending an updated patch after that

Great, thanks!

olofk:
Updated the PR now. New version is rebased against v0.23.0, puts the files in the scripts dir, removes dead code and optionally supports awatop.

PR #111, opened 2020-05-24
AXI crosspoint (axi_xp)

andreaskurth:
Adds an AXI crosspoint with homomorphous slave and master ports.

This depends on #103 (and currently uses that branch as base, which needs to be changed to `master` once #103 is merged).

## TODO
- [ ] Verify with TB similar to the one used for `axi_xbar`.


zarubaf:
Thinking about it a bit more (and dumping it here), we probably want to limit some paths through the crossbar and make it a bit simpler. For example a node with NWSE and processor output (P) with a slave and master port we probably want to disable the path from master back to slave. Relying on constant propagation imho is a bit dirty because it can very well be the case that the address map will come from a register file. What's your take on this?

accuminium:
> Thinking about it a bit more (and dumping it here), we probably want to limit some paths through the crossbar and to make it a bit simpler. For example a node with NWSE and processor output (P) with a slave and master port we probably want to disable the path from master back to slave. Relying on constant propagation imho is a bit dirty because it can very well be the case that the address map will come from a register file. What's your take on this?

I agree with your point. Would we want an additional connectivity map that defines for each slave port of the crossbar to which master port it is connected?

fabianschuiki:
I think having such a connectivity map would be very valuable! Especially since it can actually get rid of some non-trivial amount of hardware in some constellations.

accuminium:
Okay, I have added a `Connectivity` parameter that allows to define how ports are connected: https://github.com/pulp-platform/axi/pull/111/files#diff-5d663a473103fcc1d98bb5efb6115779R176-R197.

This implementation simply ties unconnected mux inputs to zero and attaches *error slaves* to unconnected demux outputs.  The synthesizer should be able to optimize the tied-off mux inputs away, but it will likely not reduce the number of demux outputs.

I also looked into more aggressive reductions of muxes and demuxes in the code depending on the connectivity map.  However, this requires remapping indices in the address map (which is not a parameter but an input port) and at demux output ports.  I think expressing this in SystemVerilog in a way compatible with most tools is very difficult, and we cannot do it without high efforts.  I would be happy to be proven wrong of course.

thommythomaso:
@fischeti. Would it be possible to update the `xp` and upstream it?

fischeti:
> @fischeti. Would it be possible to update the `xp` and upstream it?

Would be possible, yes. I can take care of that

fischeti:
I rebased the branch, pushed and cleaned up the AXI Crosspoint. It would theoretically be ready to be merged now. However, there is still no testbench to verify the crosspoint. The `tb_axi_xbar` cannot be used as is, since the crosspoint has ID remappers on the master ports, which makes the monitoring of transactions much more elaborate.

PR #116, opened 2020-06-29
AXI_XBAR pipeline

WRoenninger:
This adds a change to the `axi_demux` implementation which has the benefit that the crossing AXI busses inside of `axi_xbar` can now be pipelined.
The issue was that the `axi_demux` could forward multiple AWs to different master ports of the module. This then has the effect, that Ws are forwarded in the same sequence. However combined with an `axi_mux` this could lead to deadlocks, if the AWs where stalled for example due to pipelining. 
This PR fixes this issue by preventing AWs to be forwared to a different master port as long as there are still Ws in flight to another.

Changelog:
- `tb_axi_xbar`: Add parameters, make more configurable for ci.
- `axi_demux`: Remove FIFO between AW and W channel, is now a register plus counter.
               Prevents AWs to be emmitted downstream to a different master port as long as Ws
               are still in flight to another. This prevents deadlock, if there is stalling
               downstream.
- `axi_xbar`: Add parameter `PipelineStages` to `axi_pkg::xbar_cfg_t`. This adds `axi_multicuts`
              in the crossed connections in the xbar between the demuxes and muxes.
- `axi_pkg`: Add documentation to `xbar_cfg_t`.

### DIFF
--- a/src/axi_demux.sv
+++ b/src/axi_demux.sv
@@ -195,10 +195,10 @@ module axi_demux #(
         // Process can start handling a transaction if its `i_aw_id_counter` and `w_fifo` have
         // space in them. Further check if we could inject something on the AR channel.
         if (!aw_id_cnt_full && (w_open != {IdCounterWidth{1'b1}}) && !ar_id_cnt_full) begin
-          /// There is a valid AW vector make the id lookup and go further, if it passes.
-          /// Also stall if previous transmitted AWs still have active W's in flight.
-          /// This prevents deadlocking of the W channel. The counters are there for the
-          /// Handling of the B responses.
+          // There is a valid AW vector make the id lookup and go further, if it passes.
+          // Also stall if previous transmitted AWs still have active W's in flight.
+          // This prevents deadlocking of the W channel. The counters are there for the
+          // Handling of the B responses.
           if (slv_aw_valid &&
                 ((w_open == '0) || (w_select == slv_aw_chan_select.aw_select)) &&
                 (!aw_select_occupied || (slv_aw_chan_select.aw_select == lookup_aw_select))) begin
--- a/src/axi_demux.sv
+++ b/src/axi_demux.sv
@@ -33,7 +33,7 @@ module axi_demux #(
   parameter bit          SpillAr        = 1'b1,
   parameter bit          SpillR         = 1'b0,
   // Dependent parameters, DO NOT OVERRIDE!
-  parameter int unsigned SelectWidth    = (NoMstPorts > 32'd1) ? $clog2(NoMstPorts) : 32'd1,
+  parameter int unsigned SelectWidth    = cf_math_pkg::idx_width(NoMstPorts),
   parameter type         select_t       = logic [SelectWidth-1:0]
 ) (
   input  logic                     clk_i,
@@ -49,7 +49,8 @@ module axi_demux #(
   input  resp_t   [NoMstPorts-1:0] mst_resps_i
 );
 
-  localparam int unsigned IdCounterWidth = MaxTrans > 1 ? $clog2(MaxTrans) : 1;
+  localparam int unsigned IdCounterWidth = cf_math_pkg::idx_width(MaxTrans);
+  typedef logic [IdCounterWidth-1:0] id_cnt_t;
 
   //--------------------------------------
   // Typedefs for the FIFOs / Queues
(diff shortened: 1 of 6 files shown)


src/axi_demux.sv
```diff
@@ -234,31 +239,34 @@ module axi_demux #(
...
-      .usage_o   (                              ),
-      .data_i    ( slv_aw_chan_select.aw_select ),
-      .push_i    ( aw_push                      ), // controlled from proc_aw_chan
-      .data_o    ( w_select                     ), // where the w beat should go
-      .pop_i     ( w_fifo_pop                   )  // controlled from proc_w_chan
+
+    // This counter steers the demultiplexer of the W channel.
+    // `w_select` determines, which handshaking is connected.
+    // AWs are only forwarded, if the counter is empty, or `w_select_q` is the same as
+    // `slv_aw_chan_select.aw_select`.
```
accuminium, line 250:
```suggestion
    // This counter steers the demultiplexer of the W channel.
    // `w_select` determines which handshaking is connected.
    // AWs are only forwarded if the counter is empty or `w_select_q` is the same as
    // `slv_aw_chan_select.aw_select`.
```

But apart from this, two major questions:
1. Why is this a counter? There is only one write burst ongoing at a time, so a single-bit register should suffice, no?
2. How does this counter steer the demultiplexer of the W channel?  The demux needs to remember to which master port the W burst is going, but there is no counting involved in that, right?

src/axi_demux.sv
```diff
@@ -523,6 +531,10 @@ module axi_demux #(
     ar_stable: assert property( @(posedge clk_i) (ar_valid && !ar_ready)
                                |=> $stable(slv_ar_chan_select)) else
       $fatal(1, "slv_aw_chan_select unstable with valid set.");
+    // W counter saaertions
```
accuminium, line 534:
```suggestion
    // W counter assertions
```

src/axi_demux.sv
WRoenninger: The change still allows for multiple AWs in flight through the same master port. The counter is incremented on AW tranaction and decremented on the last flag of the W. It only stalls transmission of the AWs if they wnat to go to a different master port. The counter is for the number of active AWs to the same slave, the FF for the handshake connection of the W channel.

src/axi_pkg.sv
```diff
@@ -350,16 +350,42 @@ package axi_pkg;
...
+    /// 0: No fallthrough
+    /// 1: Fallthrough
     bit            FallThrough;
+    /// The Latency mode of the xbar. This determines if the channels on the ports have
+    /// a spill register instantiated.
+    /// Example configurations are provided with the enum `xbar_latency_e`.
     xbar_latency_e LatencyMode;
+    /// This is the number of `axi_multicut` stages instantiated in the line cross of the channels.
+    /// Having multiple stages can potentially add a large number of FFs!
+    int unsigned   PipelineStages;
```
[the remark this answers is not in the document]
WRoenninger: Removed the `Cfg` parameter struct. Its fields are now parameters in `axi_xbar`.

src/axi_pkg.sv
```diff
@@ -350,16 +350,42 @@ package axi_pkg;
 
   /// Configuration for `axi_xbar`.
   typedef struct packed {
+    /// Number of slave ports of the crossbar.
+    /// This many master modules are connected to it.
     int unsigned   NoSlvPorts;
+    /// Number of master ports of the crossbar.
+    /// This many slave modules are connected to it.
     int unsigned   NoMstPorts;
+    /// Maximum number of open transactions each master connected to the crossbar can have in
```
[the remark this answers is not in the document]
WRoenninger: Added clarification in the docstring.

WRoenninger:
Removed now the Cfg struct of `axi_xbar` and the defintion in `axi_pkg`.
This resulted in the same removal being done for `axi_lite_xbar`.

WRoenninger:
Rebased onto current master, also changed around the Bender dependencies to match current tags.

WRoenninger:
Rebased onto current master for ci fix by #129.

WRoenninger:
@accuminium There seems to be an issue with the build and deploy ci, where there is some lock failure between the (push) and (pull_request) scripts.

accuminium:
> @accuminium There seems to be an issue with the build and deploy ci, where there is some lock failure between the (push) and (pull_request) scripts.

Thanks, I also noticed that. In a PR, each of the CI jobs should actually only run once (triggered by the `pull_request` event, and not by `push`). The `push` event is relevant only outside PRs. But I have not yet spent the time to figure out how to define this condition. Please feel free to open a PR that fixes this. The workaround for now is to manually restart the `build-and-deploy` job when it fails due to a conflict with the other instance.

thommythomaso:
Replaced by #268. Closed.

PR #118, opened 2020-07-08
axi_dw_downsizer: Fix overflow of B forward FIFO

accuminium:
Before this change, `slv_resp_o.aw_ready` could be asserted even though `forward_b_beat_full`.  When that happened, the AW and W beat were not forwarded but dropped in the DWC, and upstream never got a B response for that transaction back.

To fix this, new writes in the *idle* state are only handled if `!forward_b_beat_full`.  For this, the logic handling the *idle* state has been wrapped in a [new `if` statement](https://github.com/pulp-platform/axi/pull/118/files#diff-3601b4c67184ff301f29a4fe12c0d371R699).  (The additional level of indentation confuses GitHub diff.)

### TODO 
- [ ] Changelog entry

### DIFF
--- a/src/axi_dw_downsizer.sv
+++ b/src/axi_dw_downsizer.sv
@@ -787,78 +787,80 @@ module axi_dw_downsizer #(
       w_req_d.aw_throw_error = 1'b0              ;
       w_req_d.burst_resp     = axi_pkg::RESP_OKAY;
 
-      if (slv_req_i.aw_valid && slv_req_i.aw.atop[5]) begin // ATOP with an R response
-        inject_aw_into_ar_req = 1'b1                 ;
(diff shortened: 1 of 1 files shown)


bluewww:
FYI you can just ignore whitespace changes in github diff so it looks quite clear to me
https://github.com/pulp-platform/axi/pull/118/files?diff=split&w=1

accuminium:
> FYI you can just ignore whitespace changes in github diff so it looks quite clear to me
> https://github.com/pulp-platform/axi/pull/118/files?diff=split&w=1

That's useful, thanks!

suehtamacv (approved):
Sorry for the delay in reviewing this, I missed this PR. 
Thanks for the changes, they look good to me.

accuminium:
I rebased as discussed, and CI is passing --> merging.

Issue #123, opened 2020-10-15
Long critical path through the DW Converters


suehtamacv:
Atomics are at the critical path of the DW Converters. 

Upon receiving an atomic request on the AW channel, the request is forwarded as a pseudo-AR request to one of the read up/downsizers, which will be responsible for handling the atomic responses on the R channel. The decision to forward the AW request to the read up/downsizers, the arbitration of which up/downsizer will handle the atomic request, and the logic of the up/downsizer upon receiving a new request, all that happens in a single cycle.

Pipelining this path might lead to max operating frequency improvements for the DW Converters, at the expense of lower performance for atomic operations.

accuminium:
Fixed by #124.

PR #124, opened 2020-10-16
Improve timing of the DWCs

suehtamacv:
This PR breaks the critical path of the DWCs. This is done through an extra state at the read up/downsizer FSMs, by making injected AW requests take one extra cycle. AW injection in the read up/downsizers happens in one cycle, and the evaluation of the read up/downsizing parameters happens at the next cycle.

This PR solves #123, and depends on #102.

### DIFF
--- a/src/axi_dw_downsizer.sv
+++ b/src/axi_dw_downsizer.sv
@@ -52,6 +52,8 @@ module axi_dw_downsizer #(
   import axi_pkg::aligned_addr;
   import axi_pkg::modifiable  ;
 
+  import cf_math_pkg::idx_width;
+
   // Type used to index which adapter is handling each outstanding transaction.
   localparam TranIdWidth = AxiMaxReads > 1 ? $clog2(AxiMaxReads) : 1;
   typedef logic [TranIdWidth-1:0] tran_id_t;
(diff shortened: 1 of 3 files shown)


accuminium (approved):
The changes look good to me, thanks for your work!

Can you please post the relative changes (+/- n%) in timing and area?

accuminium:
CI is currently [failing for both up- and downsizer](https://iis-git.ee.ethz.ch/akurth/axi/-/pipelines?page=1&scope=branches&ref=dwc_atomic_timing). It is also [failing for the `dwc_tbs` branch](https://iis-git.ee.ethz.ch/akurth/axi/-/pipelines?page=1&scope=branches&ref=dwc_tbs), though, so I suspect the problem is already present in that branch and not due to these changes.

Can you please look at the test runs so we have passing tests before merging this?

accuminium:
This decreases the critical path length through both upsizer and downsizer by ca. 10% while maintaining a mostly constant area.

CI is now passing, merging.

suehtamacv:
I analyzed the minimal clock period and the area of the downsizers, under the same methodology used [here](https://arxiv.org/abs/2009.05334). 

The data width converters analyzed have a critical path between 20ps and 50ps shorter than the ones before this PR. This corresponds to a 5% - 15% decrease in the critical path length. Regarding the area taken by the DWCs, they are at most 3 kGE larger, although this might be due to the higher clock frequency.

The graphs attached detail this analysis. 

Critical path length:
timing

Area: 
area

PR #127, opened 2020-11-01
Fix ipapprox deps

bluewww:
* Bump `common_cells` to  fixed version of `v1.20.0` (see https://github.com/pulp-platform/common_cells/pull/104)
* Link against `common_cells_lib`
* Fix include path. Note ipapprox has no understanding of include paths of dependencies or an option for global include paths (afaik). Anyway we have the same hack in other IPs such as fpnew and I'd rather benderize than hack around ipapprox more.

### DIFF
--- a/src_files.yml
+++ b/src_files.yml
@@ -1,6 +1,10 @@
 axi:
+  vlog_opts: [
+    -L common_cells_lib
+  ]
   incdirs:
     - include
+    - ../../common_cells/include
   files:
     # Source files grouped in levels. Files in level 0 have no dependencies on files in this
     # package. Files in level 1 only depend on files in level 0, files in level 2 on files in
--- a/Bender.yml
+++ b/Bender.yml
@@ -8,7 +8,7 @@ package:
     - "Wolfgang Roenninger <wroennin@ethz.ch>"
 
 dependencies:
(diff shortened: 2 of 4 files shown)


src_files.yml
```diff
@@ -1,6 +1,10 @@
 axi:
+  vlog_opts: [
+    -L common_cells_lib
+  ]
   incdirs:
     - include
+    - ../../common_cells/include
```
accuminium, line 7: Is the common convention of including directories from other dependencies in ipapprox two levels up (`../..`)? Or will this only work for a specific project?

accuminium:
Once https://github.com/pulp-platform/common_cells/pull/103 has been resolved and included in a release, please also update the version of `common_cells` in `Bender.yml`, so we stay consistent.

src_files.yml
bluewww: Normally, you nest axi related ips into a top level axi directory. Common_cells is then on the same level as that top level axi directory. This is at least done in the projects that (still) use ipapprox (pulp, pulpissimo and our internal forks thereof).

accuminium: Okay, then I am fine with it.

bluewww:
Depends on https://github.com/pulp-platform/common_cells/pull/104

accuminium (approved):
LGTM, thanks!

PR #131, opened 2020-12-16
Xsim

olofk:
This introduces a bunch of changes needed for compatibility with Vivado XSIM. Also adds a FuseSoC core descirption file

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   `AXI_BUS` (et al.) above.
 
 ### Fixed
+- `axi_demux`: Break combinatorial simulation loop.
 - `axi_xbar`: Improve compatibility with vsim version 10.6c (and earlier) by introducing a
   workaround for a tool limitation (#133).
 - `tb_axi_lite_regs`: Removed superfluous hardcoded assertion.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## Unreleased
 
 ### Added
+- `axi_xbar`: Add parameter to disable support for atomic operations (`ATOPs`).
 
 ### Changed
 - `AXI_BUS`, `AXI_BUS_ASYNC`, `AXI_BUS_DV`, `AXI_LITE`, and `AXI_LITE_DV`: Change type of every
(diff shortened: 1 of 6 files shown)


accuminium:
Thanks for these changes and fixes @olofk, much appreciated!

The changes essentially LGTM, but regarding adding a `NDEBUG` define, I wonder if this is a kind-of canonical or standard way of disabling assertions in simulators that do not support them?  I did not find anything in the SystemVerilog standard.  A `NDEBUG` define at least may be intuitive for those familiar with [the C standard library](https://en.cppreference.com/w/c/error/assert).

The background is that int the long run, we would like to switch to the [assert macros provided by lowRISC/OpenTitan](https://github.com/lowRISC/opentitan/blob/master/hw/ip/prim/rtl/prim_assert.sv), which again would break compatibility with xsim.  In fact, (an older version of) those macros are already in `assertions.svh` of our `common_cells`, for which [you submitted a PR as well](https://github.com/pulp-platform/common_cells/pull/107).  However, if we upgrade `assertions.svh` to the upstream `prim_assert.sv`, the xsim fix is lost again.  So how about fixing this upstream by adding an `` `ifdef XSIM`` to the [list of special cases in `prim_assert.sv`](https://github.com/lowRISC/opentitan/blob/2b99fa4339dc23f67b9401a448c3fe02cd89fa1b/hw/ip/prim/rtl/prim_assert.sv#L72-L82) (or `NDEBUG`, depending on what upstream thinks)?

imphil:
From an OpenTitan perspective, a ifdef XSIM in prim_assert.sv (https://github.com/lowRISC/opentitan/blob/master/hw/ip/prim/rtl/prim_assert.sv#L72) is preferred. There is unfortunately no tool-agnostic way to do that (every tool is expected to support all of SystemVerilog in general).

olofk:
Thanks @imphil!
@accuminium, in the light of this I will amend the patches to use an `XSIM` ifdef rather than`NDEBUG`. Ok with you?

zarubaf:
@olofk Could we do the same on the `common_cells` repo (again) so that everything is inline? Very much appreciated!

accuminium:
Perfect, thanks all!

olofk:
@zarubaf Yes, I think that makes sense. Will come up with a patch shortly

olofk:
@accuminium @zarubaf I replace the NDEBUG define with an XSIM version now and did the same to common_cells.

For the record my experience tells me it's not a very good idea to do things like this based on tool names since tools tend to add/remove features over time and have certain versions where things are broken so that it's better to have define/paramter names that are more descriptive and let the build system figure out when to set them. But I guess I should take this to the OpenTitan devs instead

accuminium:
Thanks @olofk! I have rebased your branch on the current `master`, updated the Changelog, slightly reworded some commit messages to comply with our conventions, and added a commit with the assertions for `axi_demux` that you suggested. Let's get this merged :+1:

> For the record my experience tells me it's not a very good idea to do things like this based on tool names since tools tend to add/remove features over time and have certain versions where things are broken. It's better to have define/paramter names that are more descriptive and let the build system figure out when to set them. But I guess I should take this to the OpenTitan devs instead

I see your point and I agree that you should discuss this with the OpenTitan devs.

olofk:
Awesome! This means we can run SweRVolf on xsim without any downstream patches. Thanks

PR #135, opened 2021-01-11
Fix `randomize()` for class variables

### LINKED ISSUE #134: Simulation Error
What simulation tool to use ? I use VCS there are some compile error:
Error-[UARC] Unsupported argument to randomize call
/nfs54/project/spiderman/zhuzhiqi/git/axi_pulp/axi/src/axi_test.sv, 1107
"w_beat"
  Arg #1 of std::randomize "w_beat" is not integral or enum or array of 
  integral or enums
.......

accuminium:
Fixes #134.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


skokvermon:
unfortunately this doesn't work in vivado simulator 2021.2

andreaskurth:
Thanks for reporting; let's solve this as discussed in #226.

Issue #128, opened 2020-11-12
VCS compilation issue for AXI lite modules 


saadkhalil283:
Please see the following error message shown by vcs

Error-[SV-UIP] Unconnected interface port
axi_lite_demux.sv, 421
"slv"
  The port 'slv' of top-level module 'axi_lite_demux_intf' whose type is 
  interface 'AXI_LITE' is left unconnected. It is illegal to leave the 
  interface ports unconnected.
  Please make sure that all the interface ports are connected.

accuminium:
> The port 'slv' of **top-level module** 'axi_lite_demux_intf' whose type is interface 'AXI_LITE' is left unconnected.

Apparently, VCS does not support SystemVerilog interfaces in top-level modules (which would be a restriction shared with many other tools).  However, `axi_lite_demux` is not designed to be a top-level module.  Are you sure that you want it to be the top-level module?  If not, you should change the top-level module in VCS.

saadkhalil283:
Yes, _axi_lite_demux_ is not the top module. _tb_axi_lite_xbar_ is the top module. The problem is while compiling through the hierarchy when VCS compiles _axi_lite_demux_intf_, the module contains an interface instance of _AXI_LITE_ for both its **Master** and **Slave** modports. That's where it raises this error. Same error is also raised in _axi_lite_mux_intf_ and _axi_to_axi_lite_intf_.

accuminium:
I tried to reproduce the error you reported but could not.  I was able to simulate `tb_axi_lite_xbar` in VCS 2020.03. Here is what I did:
1.  I checked out the newly-created [`vcs-improve-compatibility` branch](https://github.com/pulp-platform/axi/tree/vcs-improve-compatibility) (more on that below).
2.  I created a `build` directory and cd'ed into it.
3.  I used [Bender 0.21.0](https://github.com/fabianschuiki/bender/tree/v0.21.0) to generate a VCS compile script:
    ```sh
    bender script vcs -t test --vlog-arg='-assert svaext' --vlog-arg='-timescale=1ps/1ps' > compile_vcs.sh
    ```
4.  I made the `compile_vcs.sh` script executable and ran it.
5.  I compiled the `simv` binary for `tb_axi_lite_xbar` with
    ```sh
    vcs -full64 tb_axi_lite_xbar -debug_access+all -notice
    ```
6.  I ran the `simv` binary and saw the TB successfully run to completion.

[...]

PR #136, opened 2021-01-11
Improve compatibility with VCS (1/2)

accuminium:
These are a couple of backwards-compatible changes to improve compatibility with VCS.


accuminium:
See #128.

PR #139, opened 2021-01-12
Fix issue 122

### LINKED ISSUE #122: tb_axi_lite_regs: Random load collides with expected response
Hey!
In specific seed they value of `rand_load` in tests `tb_axi_lite_regs` causes assertion fail:
```
KERNEL: 1500000 Lite Master> Recieved read response from ADDR: 00000000 DATA: de01beef RESP: 0
KERNEL: Fatal Error: /home/tester/gitlab/axi/test/tb_axi_lite_regs.sv (124): Data is unexpected, should be axi_data_t'(64'hDEADBEEFDEADBEEF).
```
If we change seed, test works fine.

WRoenninger:
This fixes #122.

The tb performs a known write followed by a read at the begining of simulation to a writeable address. The hardcoded data read assertion can be safely removed as the read data is checked by `proc_check_read_data` at the same time against the dynamic model.

Also added some more parameters in the vsim run script for the tb.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - `axi_xbar`: Improve compatibility with vsim version 10.6c (and earlier) by introducing a
   workaround for a tool limitation (#133).
+- `tb_axi_lite_regs`: Removed superfluous hardcoded assertion.
 
 
 ## 0.24.2 - 2021-01-11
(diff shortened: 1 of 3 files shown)


accuminium (approved):
Great, thanks!

accuminium:
Can you please rebase on `master`, so we can merge this?

WRoenninger:
Rebased and added entry to CHANGELOG.

accuminium:
> Rebased and added entry to CHANGELOG.

In general, we do not add TB changes to the Changelog, because they are not part of the public "interface" of the repository.  You can keep this one, however.

PR #145, opened 2021-01-15
assign.svh: Add macros for assigning between structs and generalize `AXI_TO_*` to any LHS and RHS

accuminium:


fabianschuiki (approved):
This is an excellent change! Makes things much more streamlined and removes a lot of duplication :+1: :100:

PR #146, opened 2021-01-18
axi_id_remap: Check the width of the whole `w` channel

SamuelRiedel:
Very minor fix. Since the `w` channel has no `id` and is directly forwarded from the input to the output, we can check that the width of the whole channel matches at the input and output. Thereby, the `data`, `strb`, and `user` signals have to match.

Would you prefer checking the `w.data` and the `w.user` width separately to eliminate the case where both signals' width mismatch, but the total width is the same for both ports?


accuminium:
> Would you prefer checking the `w.data` and the `w.user` width separately to eliminate the case where both signals' width mismatch, but the total width is the same for both ports?

Yes please!

accuminium (approved):
LGTM, thanks!

PR #147, opened 2021-01-19
typedef: Provide macros for generating all types

zarubaf:
This can be used whenever the user is not interested in "precise" control of the naming of the individual channels.


accuminium (approved):
LGTM, thanks!

PR #148, opened 2021-01-19
axi_test: Randomize `qos`

zarubaf:
For the random AXI master it can be interesting to also randomize the
qos signal, especially in view of achieving any code coverage on logic
dependent on this signal.

### DIFF
--- a/src/axi_test.sv
+++ b/src/axi_test.sv
@@ -236,7 +236,7 @@ package axi_test;
     logic               ax_lock   = '0;
     logic [3:0]         ax_cache  = '0;
     logic [2:0]         ax_prot   = '0;
-    logic [3:0]         ax_qos    = '0;
+    rand logic [3:0]    ax_qos    = '0;
     logic [3:0]         ax_region = '0;
     logic [5:0]         ax_atop   = '0; // Only defined on the AW channel.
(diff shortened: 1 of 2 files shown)


src/axi_test.sv
```diff
@@ -835,7 +836,9 @@ package axi_test;
 
       ax_beat.ax_addr = addr;
       rand_success = std::randomize(id); assert(rand_success);
+      assert(std::randomize(qos));
```
accuminium, line 839:
```suggestion
      rand_success = std::randomize(qos); assert(rand_success);
```
Assertions can be disabled, in which case the arguments to `assert()` are not evaluated. To prevent this, asserting the success of randomization should be done in two steps.

src/axi_test.sv
```diff
@@ -835,7 +836,9 @@ package axi_test;
...
       ax_beat.ax_id = id;
+      ax_beat.ax_qos = qos;
```
accuminium, line 841: I suggest to also randomize `ax_qos` in the `rand_atop_burst` function (just below).

accuminium (changes requested):
Thanks for these changes! I agree with the idea, but I have two change requests.

accuminium (approved):

PR #149, opened 2021-01-19
axi_test: Optionally randomize responses

zarubaf:
This patch makes it optional for a random slave to randomize the success
of any accesses. The default is off, so any existing code should be
still compatible.

Again, useful for achieving code coverage on error conditions of any
modules.

### DIFF
--- a/src/axi_test.sv
+++ b/src/axi_test.sv
@@ -1177,6 +1177,7 @@ package axi_test;
     // Stimuli application and test time
     parameter time  TA = 0ps,
     parameter time  TT = 0ps,
+    parameter bit   RAND_RESP = 0,
     // Upper and lower bounds on wait cycles on Ax, W, and resp (R and B) channels
     parameter int   AX_MIN_WAIT_CYCLES = 0,
     parameter int   AX_MAX_WAIT_CYCLES = 100,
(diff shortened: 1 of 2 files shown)


src/axi_test.sv
```diff
@@ -1244,6 +1245,8 @@ package axi_test;
         ar_beat = ar_queue.peek();
         rand_success = r_beat.randomize(); assert(rand_success);
         r_beat.r_id = ar_beat.ax_id;
+        if (RAND_RESP && !ar_beat.ax_atop[5])
```
accuminium, line 1248:
```suggestion
        if (RAND_RESP && !ar_beat.ax_atop[axi_pkg::ATOP_R_RESP])
```
I think this improves the expressiveness of the condition

src/axi_test.sv
```diff
@@ -1294,6 +1297,8 @@ package axi_test;
         aw_beat = aw_queue.pop_front();
         rand_success = b_beat.randomize(); assert(rand_success);
         b_beat.b_id = aw_beat.ax_id;
+        if (RAND_RESP && !aw_beat.ax_atop[5])
```
accuminium, line 1300:
```suggestion
        if (RAND_RESP && !aw_beat.ax_atop[axi_pkg::ATOP_R_RESP])
```
as above

accuminium (changes requested):
Thanks, I agree with the idea. Two small change requests

zarubaf:
Agree.

accuminium (approved):

Issue #153, opened 2021-01-20
Harmonize ports and parameters


accuminium:
Let us collect the changes required to harmonize ports and parameters and to minimize incompatibilities with EDA tools.  Those changes will be breaking (as in "backwards-incompatible"), so let us make sure we get them right.  **This is currently a draft** and contributors are kindly asked to comment.

Preamble: The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [BCP 14](https://tools.ietf.org/html/bcp14) [[RFC2119]](https://tools.ietf.org/html/rfc2119) [[RFC8174]](https://tools.ietf.org/html/rfc8174) when, and only when, they appear in all capitals, as shown here.

### Legal Types

Every `parameter` of a synthesizable `module` MUST be either:
(a) a `type`, or
(b) a (vector of) one of the following SystemVerilog types:
  - `bit` or `logic`, which MAY be `signed` but by default are `unsigned`, or
  - `byte`, `shortint`, `int`, or `longint`, which MUST be followed by `unsigned` or `signed`, or

(c) a `typedef`ed type of one of the types in (b).

In particular, `struct`s and `string`s MUST NOT be used as `parameter` of a `module`.

Rationale: Many tools do not properly implement complex types for parameters.

For non-synthesizable `module`s and `class`es, the key words MUST and MUST NOT in this section are relaxed to SHOULD and SHOULD NOT, respectively.  (In particular, testbench `module`s MAY use `time` and `string` parameters.)

### Signedness

If an integer parameter (i.e., `byte`, `shortint`, `int`, or `longint`) is not supposed to take negative values, it MUST be declared `unsigned` instead of `signed`.

### Default Value

Every `parameter` MUST have a default value.

If possible, the default value SHOULD be a *null* value that is outside the legal range of the parameter (e.g., a signal width of zero).  In this case, the `module` SHOULD contain an assertion to ensure that the parameter is set to a value other than the *null* value at instantiation.

Rationale: Many tools require `parameter`s to have a default value, but in many cases a `parameter` that is not set at instantiation indicates an error that should be detected. 

### Derived Parameters

The parameter list of a `module` MUST NOT contain `localparam`s.  Rationale: Unsupported by some tools.

[...]

PR #157, opened 2021-01-26
WIP: AXI_XBAR parameters

WRoenninger:
This is a first WIP to update `axi_xbar` according to #153.

I also added and adapted the documentation that was in `doc/axi_xbar.md` for inline for auto generation.
The doc file therefore should probably be removed when this is merged to prevent documentation duplication.

* Renamed ports and paramters of `axi_xbar`, added inline documentation.
* Added parameter sweep to the xbar portion of `scripts/run_vsim.sh`
* Renamed TB parameters / added appropriate macros for assign
  * Changed wrong transfer warnings in `tb_axi_xbar_pkg` to errors

The `axi_pkg::xbar_cfg_t` struct is still in as it is currently used in `axi_lite_xbar` also and can be removed when its ports are updated.
  
Edit: Also when looking at https://pulp-platform.github.io/axi/xbar_parameters/module.axi_xbar.html the link to doc/axi_xbar.png is currently broken. How do I link properly tho the image in the inline documentation?


accuminium:
Very nice, thanks already!

> Edit: Also when looking at https://pulp-platform.github.io/axi/xbar_parameters/module.axi_xbar.html the link to doc/axi_xbar.png is currently broken. How do I link properly tho the image in the inline documentation?

You need to move `axi_xbar.png` into `docs/` folder (does not yet exist on `master`). You can then specify a path relative to that folder.

src/axi_xbar.sv
```diff
@@ -13,311 +13,601 @@
...
+  parameter int unsigned AddrWidth = 32'd0,
+  /// AXI4+ATOP data field width.
+  parameter int unsigned DataWidth = 32'd0,
+  /// AXI4+ATOP user field width.
+  parameter int unsigned UserWidth = 32'd0,
+  /// Maximum number of open transactions each master connected to the crossbar can have
+  /// [in flight](doc/#in-flight) at the same time.
+  parameter int unsigned SlvPortMaxTxns = 32'd0,
+  /// Maximum number of open transactions each slave connected to the crossbar can have
+  /// [in flight](../doc#in-flight) per ID at the same time.
```
accuminium, line 175: Either this link or the one above is off, no?

src/axi_xbar.sv
```diff
@@ -13,311 +13,601 @@
...
+  /// Setting this to small values leads to less area, however on an increased stalling rate
+  /// due to ID collisions.
+  parameter int unsigned SlvPortIdWidthUsed = 32'd0,
+  /// AXI4+ATOP address field width.
+  parameter int unsigned AddrWidth = 32'd0,
+  /// AXI4+ATOP data field width.
+  parameter int unsigned DataWidth = 32'd0,
+  /// AXI4+ATOP user field width.
+  parameter int unsigned UserWidth = 32'd0,
+  /// Maximum number of open transactions each master connected to the crossbar can have
```
accuminium, line 171:
```suggestion
  /// Maximum number of open transactions each slave port of the crossbar can have
```

src/axi_xbar.sv
```diff
@@ -13,311 +13,601 @@
...
+  /// [in flight](doc/#in-flight) at the same time.
+  parameter int unsigned SlvPortMaxTxns = 32'd0,
+  /// Maximum number of open transactions each slave connected to the crossbar can have
```
accuminium, line 174:
```suggestion
  /// Maximum number of open transactions each master port the crossbar can have
```

src/axi_xbar.sv
```diff
@@ -13,311 +13,601 @@
...
+/// This is the interface wrapper for `axi_xbar`. Ports and parameters are analog to `axi_xbar`,
+/// see [`axi_xbar` documentation](module.axi_xbar).
+/// The AXI4+ATOP master and slave ports are structured here as interfaces.
+///
+/// The indexing of the master and slave port interface arrays is big-endian.
 module axi_xbar_intf #(
-  parameter int unsigned AXI_USER_WIDTH =  0,
-  parameter axi_pkg::xbar_cfg_t Cfg     = '0,
-  parameter type rule_t                 = axi_pkg::xbar_rule_64_t
+  /// See [`axi_xbar`](module.axi_xbar).
```
accuminium, line 491: I do not see the value of having this line above every parameter and port. We already state in the module description that its parameters and ports are described elsewhere. I suggest to remove those lines.

andreaskurth (approved):
Essentially LGTM in a high-level review, thanks a lot!

src/axi_xbar.sv
```diff
@@ -13,311 +13,601 @@
...
+/// different master ports of the multiplexer exclude each other because the order is given by the
+/// arbitration tree of the AW channel.
+///
+/// The second and third criterion are inherent to the AXI protocol:  For (2), the valid signal has
+/// to be held high until the ready signal goes high.  For (3), AXI does not allow interleaving of
+/// W beats and requires W bursts to be in the same order as AW beats.
+///
+/// The fourth criterion is thus the only one that can be broken to prevent deadlocks.  However,
+/// inserting a spill register between a master port of the [`axi_demux`](module.axi_demux) and a
+/// slave port of the [[`axi_mux`](module.axi_mux) can lead to a circular dependency inside the
```
accuminium, line 128:
```suggestion
/// slave port of the [`axi_mux`](module.axi_mux) can lead to a circular dependency inside the
```

src/axi_xbar.sv
```diff
@@ -13,311 +13,601 @@
...
 // - Florian Zaruba <zarubaf@iis.ee.ethz.ch>
 
-// axi_xbar: Fully-connected AXI4+ATOP crossbar with an arbitrary number of slave and master ports.
-// See `doc/axi_xbar.md` for the documentation, including the definition of parameters and ports.
+`include "axi/typedef.svh"
+
+/// # AXI4+ATOP Fully-Connected Crossbar
+///
+/// `axi_xbar` is a fully-connected crossbar that implements the full AXI4 specification plus atomic
+/// operations (ATOPs) from AXI5 (E1.1).
```
accuminium, line 21:
```suggestion
/// Fully-connected crossbar that implements AXI4 plus atomic operations (ATOPs) from AXI5 (E1.1).
```

The first line appears in the list of modules. It should therefore contain a concise description of the module.

src/axi_xbar.sv
```diff
@@ -13,311 +13,601 @@
...
+  ///
+  /// Example types are provided in [`axi_pkg`](package.axi_pkg).
+  ///
+  /// Required struct fields:
+  ///
+  /// typedef struct packed {
+  ///   int unsigned idx;
+  ///   axi_addr_t   start_addr;
+  ///   axi_addr_t   end_addr;
+  /// } rule_t;
```
accuminium, line 231:
```suggestion
  /// ```systemverilog
  /// typedef struct packed {
  ///   int unsigned idx;
  ///   axi_addr_t   start_addr;
  ///   axi_addr_t   end_addr;
  /// } rule_t;
  /// ```
```

src/axi_xbar.sv
```diff
@@ -13,311 +13,601 @@
...
+/// One address map is shared by all master ports.  The *address map* contains an arbitrary number
+/// of rules (but at least one).  Each *rule* maps one address range to one master port.  Multiple
+/// rules can map to the same master port.  The address ranges of two rules may overlap: in case two
+/// address ranges overlap, the rule at the higher (more significant) position in the address map
+/// prevails.
+///
+/// Each address range includes the start address but does **not** include the end address.
+/// That is, an address *matches* an address range if and only if
+/// ```
+///     addr >= start_addr && addr < end_addr
```
accuminium, line 52:
```suggestion
/// addr >= start_addr && addr < end_addr
```
Indentation is not required inside a backtick block.

src/axi_xbar.sv
```diff
@@ -13,311 +13,601 @@
...
+  /// typedef struct packed {
+  ///   int unsigned idx;
+  ///   axi_addr_t   start_addr;
+  ///   axi_addr_t   end_addr;
+  /// } rule_t;
+  parameter type rule_t = axi_pkg::xbar_rule_64_t,
+  /// Dependent parameter, do **not** override!
+  /// Width of the index specifying a master port.
+  parameter int unsigned DefaultMstPortIdxWidth = cf_math_pkg::idx_width(NumMstPorts),
+  /// Dependent parameter, do **not**parameter override!
```
accuminium, line 236:
```suggestion
  /// Dependent parameter, do **not** override!
```

src/axi_xbar.sv
```diff
@@ -13,311 +13,601 @@
...
-  input  logic                                                       rst_ni,
-  input  logic                                                       test_i,
-  input  slv_req_t  [Cfg.NoSlvPorts-1:0]                             slv_ports_req_i,
-  output slv_resp_t [Cfg.NoSlvPorts-1:0]                             slv_ports_resp_o,
-  output mst_req_t  [Cfg.NoMstPorts-1:0]                             mst_ports_req_o,
-  input  mst_resp_t [Cfg.NoMstPorts-1:0]                             mst_ports_resp_i,
-  input  rule_t     [Cfg.NoAddrRules-1:0]                            addr_map_i,
-  input  logic      [Cfg.NoSlvPorts-1:0]                             en_default_mst_port_i,
-  input  logic      [Cfg.NoSlvPorts-1:0][$clog2(Cfg.NoMstPorts)-1:0] default_mst_port_i
+  /// Clock, positive edge triggered.
```
accuminium, line 240:
```suggestion
  /// Clock, rising edge triggered.
```

src/axi_xbar.sv
```diff
@@ -13,311 +13,601 @@
...
+  input  rule_t [NumAddrRules-1:0] addr_map_i,
+  /// Enables a default master port for each slave port. When this is enabled unmapped
+  /// transactions get issued at the master port given by `default_mst_port_i`.
+  /// Each bit index corresponds to the index of a master port and is ordered little-endian (downto).
+  ///
+  /// When not used, tie to `'0`.
+  input  logic [NumSlvPorts-1:0] en_default_mst_port_i,
+  /// For each slave port the default index where the transaction should be routed, if
+  /// for this slave port the default index functionality is enabled by setting the
+  /// bit `en_default_mst_ports_i[slave_port_idx]` to `'1`.
```
accuminium, line 268:
```suggestion
  /// bit `en_default_mst_ports_i[slave_port_idx]` to `1'b1`.
```

src/axi_pkg.sv
```diff
@@ -393,16 +393,39 @@ package axi_pkg;
...
     int unsigned   AxiIdUsedSlvPorts;
+    /// AXI4+ATOP address field width.
     int unsigned   AxiAddrWidth;
+    /// AXI4+ATOP data field width.
     int unsigned   AxiDataWidth;
+    /// The number of address rules defined for routing of the transactions.
+    /// Each master port can have multiple rules, should have however at least one.
+    /// If a transaction can not be routed the xbar will answer with an `axi_pkg::RESP_DECERR`.
     int unsigned   NoAddrRules;
   } xbar_cfg_t;
```
accuminium, line 430: This `struct` can be removed, right?

src/axi_xbar.sv
WRoenninger: Seems that both are not working. So the referenced document also has to go into the docs/ folder?

src/axi_xbar.sv
WRoenninger: Removed them, however left the two lines for the slave and master port interface arrays as the indexing changes.

src/axi_pkg.sv
WRoenninger: Still used in `àxi_lite_xbar`.

src/axi_xbar.sv
accuminium: Ah, I see; because the link should go to https://github.com/pulp-platform/axi/tree/master/doc#in-flight, which is part of the repository, not of the documentation tree.

What we probably want is to add this information to the index document of the documentation. That is, https://pulp-platform.github.io/axi/master/index.html should contain a *Terminology* section, to which we can then link. The relative reference to the definition in the index would be `index#in-flight`.

WRoenninger:
Added the `docs/` folder and copied the `axi_xbar.svg/.png` to it.
Are the now redundant files for the `axi_xabr` documentation save to delete from the `doc/` folder with this PR?

accuminium:
> Are the now redundant files for the `axi_xabr` documentation save to delete from the `doc/` folder with this PR?

Yes

WRoenninger:
Seems however that the link to the .png in line `28` to `docs/axi_xbar.png` is still broken. How do I specify this properly, or does it only show up on master?

WRoenninger:
Is there a possibility to check the pipeline status of a failing gitlab-ci? Just to check if the ci failed on a stall due to a functional failing module, or that the timeout time was not long enough and the ci was still simulating.

accuminium:
> Seems however that the link to the .png in line `28` to `docs/axi_xbar.png` is still broken. How do I specify this properly, or does it only show up on master?

You need to remove the `docs/` prefix; the relative URL is simply `axi_xbar.png`.

> Is there a possibility to check the pipeline status of a failing gitlab-ci? Just to check if the ci failed on a stall due to a functional failing module, or that the timeout time was not long enough and the ci was still simulating.

Only if you have access to the mirrored repository on which CI runs. For this branch, the timeout was too low, because the crossbar simulated for more than 3 hours.

WRoenninger:
> Why is that?

The simulation script currently has 5 different parameters which each get checked for two values in the script. This equates to 32 simulation runs. In the tb each instantiated rand_master completes 100 read and 100 write bursts. However the wider tb configurations run quite slow as there is a lot going on (e.g. 6 slv_ports, 9 mst_ports). Adding new the `EnableAtops` parameter doubled the runtime. 

I could tune down the numbers though, e.g. to less `slv_ports`. I suspect they tend to slow down the simulation quite a bit.

Would it be reasonable to split the script for `axi_xbar`? So that there is not only one `axi_xbar` thread in the ci, but multiple. However I think the simulate script is currently not set up for this in a scalable / non-hacky manner. 
Also adding the internal pipeline from #116 will again double the number of simulations to be run by the ci.

Also similar issue will be in #33, there are at least 5 parameters which should be checked in range by the ci.

WRoenninger:
Just seen: As is `scripts/axi_intercon_gen.py` will be broken after implementing these changes.

PR #158, opened 2021-01-28
Update `common_verification` dependency

micprog:
This PR updates the `common_verification` dependency.
This includes renaming parameters of called modules for compatibility, removing `timeunit/timeprecision` arguments, and updating the `common_cells` dependency with the same updates, avoiding version conflicts.


accuminium (approved):
LGTM. Thanks for these changes, @micprog!

PR #159, opened 2021-01-29
Add Verilator script

zarubaf:
There was a need to rename the `synth_bench` to `axi_synth_bench` as it would collide with `common_cells` `synth_bench`. I will do the same for the `common_cells` but I figured it would be good to fix this in any case (thanks for the good namespacing support SV :-)).

Furthermore, there it contains a small fix with the unpacked array literals.

Verilator 4.100 works fine with the synthesis testbench but we got quite some warnings. I think we should go through them and figure out which one we want to waive. Then we can also check, that new additions to the repo don't cause any new warnings.

The next step will be to add Verilator to the CI regression, first to check for incompatible constructs, second to check the lint warnings.

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -55,7 +55,7 @@ sources:
 
   - target: synth_test
     files:
-      - test/synth_bench.sv
+      - test/axi_synth_bench.sv
 
   - target: simulation
     files:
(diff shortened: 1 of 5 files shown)


accuminium (approved):
Thanks, LGTM!

accuminium:
Do you prefer to merge this now, or would you like to clean the lint warnings first and add a CI job?

zarubaf:
I would prefer merging this now, just to keep the changes incremental and small. I would then go for two/three new PRs:

- Adding pass/fail to the CI
- Fixing Lint warnings
- Adding warnings to pass/fail criterium

PR #163, opened 2021-01-30
axi_up/downsizer: remove forward reference

LucaZulberti:
DesignCompiler 2019.12 reports it as a possible error in future versions.


src/axi_dw_upsizer.sv
```diff
@@ -241,6 +241,19 @@ module axi_dw_upsizer #(
     R_INCR_UPSIZE
   } r_state_e;
 
+  /* Write-related type, but w_req_q is referenced from Read logic */
```
accuminium, line 244:
```suggestion
  // Write-related type, but w_req_q is referenced from Read logic
```

accuminium, line 244:
```suggestion
  // Write-related type, but w_req_q is referenced from Read logic
```

accuminium (changes requested):
Thanks for catching and fixing this, @LucaZulberti! LGTM, only a small nitpick.

LucaZulberti:
Changed ;-)

accuminium (approved):
Thanks!

suehtamacv (approved):

PR #164, opened 2021-02-01
axi_lite_to_apb: Make pipeline registers optional 

accuminium:
Rationale: The conversion from AXI-Lite to APB is fairly simple both for
requests and responses.  Thus, unless this needs to be clocked at very
high frequencies, it should be possible without a pipeline register.
This saves one cycle of latency for each request and one for each
response.  Since latency is presumably more critical for this conversion
that achieving maximum frequency, the default is changed to not insert
pipeline registers.

The registers cannot be completely omitted, however, because the
implementation relies on a default-ready behavior both for requests and
responses.  Fall-through registers, which are designed exactly for
adding the default-ready behavior, are thus instantiated when pipelining
is disabled.


WRoenninger (approved):
LGTM!

Update of the ports and parameters to #153 will be done separately?

Yes, implementation needs the default readyness / was designed with that behaviour in mind.

accuminium:
> LGTM!
> 
> Yes, implementation needs the default readyness / was designed with that behaviour in mind.

Great, thanks for the confirmation!

> Update of the ports and parameters to #153 will be done separately?

Yes.

PR #168, opened 2021-04-08
axi_cdc: Modularize into source and destination side

luca-valente:
1. The implementation of `axi_cdc` is replaced with an instance of `axi_cdc_src` connected to an instance of `axi_cdc_dst`.
2. `axi_cdc_*` use `parameter type aw_chan_t = logic` (and similar for the other channels), as is currently done for `axi_cdc`.  
3. The main implementation of `axi_cdc_src` and `axi_cdc_dst` use `struct`s as ports and format parameter names as `lower_snake_case`    for types and `UpperCamelCase` for numbers.  
4. Added also the async interfaces definition

### DIFF
--- a/.ci/Memora.yml
+++ b/.ci/Memora.yml
@@ -55,6 +55,8 @@ artifacts:
       - src/axi_pkg.sv
       - src/axi_intf.sv
       - src/axi_test.sv
+      - src/axi_cdc_dst.sv
+      - src/axi_cdc_src.sv
       - src/axi_cdc.sv
       - test/tb_axi_cdc.sv
     outputs:
--- a/Bender.yml
+++ b/Bender.yml
@@ -25,7 +25,8 @@ sources:
   # Level 2
   - src/axi_atop_filter.sv
   - src/axi_burst_splitter.sv
-  - src/axi_cdc.sv
+  - src/axi_cdc_dst.sv
+  - src/axi_cdc_src.sv
   - src/axi_cut.sv
   - src/axi_delayer.sv
   - src/axi_demux.sv
@@ -45,6 +46,7 @@ sources:
   - src/axi_mux.sv
   - src/axi_serializer.sv
   # Level 3
+  - src/axi_cdc.sv
   - src/axi_err_slv.sv
   - src/axi_dw_converter.sv
   - src/axi_multicut.sv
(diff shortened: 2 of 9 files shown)


accuminium:
Thanks for these changes, @luca-valente, much appreciated!

I am going through the changes and will make my edits directly to your commit.  So far I agree with the essence of the changes.

However, I have one major concern regarding the `isolate_i` input.  If I am interpreting your intention correctly, that signal should disable transactions on all channels while it is high.  Is this correct?  If so, simply ANDing either the ready or the valid signal of each channel with `~isolate_i` will unfortunately not work. For example, consider the AW channel Gray CDC FIFO in `axi_cdc_dst`: `~isolate_i` is currently ANDed with `dst_resp_i.aw_ready`.  So if `isolate_i` is high, the handshake does not happen for the FIFO, which is as intended. However, the slave port that drives `dst_resp_i.aw_ready` does not know this, and it assumes that the handshake happens when it sets `dst_resp_i.aw_ready` high while `dst_req_o.aw_valid` is high. This mismatch will corrupt data and protocol state. One might be tempted to resolve this mismatch by also ANDing `dst_req_o.aw_valid` with `~isolate_i`. However, this would violate the stability guarantee of the valid signal in AXI's handshaking rules. Thus, properly isolating an AXI interface is more complex than ANDing a few handshake signals. Fortunately, we already have the [`axi_isolate` module](https://github.com/pulp-platform/axi/blob/v0.27.1/src/axi_isolate.sv), which does exactly this (independent of a CDC).  So I suggest to remove the `isolate_i` functionality and use the `axi_isolate` module instead.  If you agree, I can make this change as part of my edits, so we don't have to resolve merge conflicts.

luca-valente:
Hi @accuminium !  To me it is okay to remove the `isolate_i`. Maybe @meggiman has some opinion on this since in Pulpissimo `isolate` is as static signal set high so that the `axi_cdc_fifos` are removed by Synopsys. Maybe he has already been using another way to do so.

meggiman:
I aggree with @accuminium  on the isolate_i signal. For the integration in pulpissimo I think we can just remove the CDC IP altogether and instead parametrically instantiate e.g. an AXI error slave to indicate that there is no cluster to communicate with.

PR #172, opened 2021-04-22
axi_xbar and axi_demux: Add support for unique IDs

accuminium:
Add support for unique IDs by adding a `UniqueIds` parameter to `axi_xbar` and `axi_demux`.  If you can guarantee that the ID of each transaction is always unique among all in-flight transactions in the same direction, setting the `UniqueIds` parameter to `1'b1` simplifies the demultiplexer (see documentation of `axi_demux` for details).

This change is backward-compatible on `axi_demux` (because the default value of the new parameter is `1'b0`).  As `axi_xbar` is configured with the `xbar_cfg_t` `struct`, this change is *not backward-compatible* for `axi_xbar` (except for `xbar_cfg_t`s initialized with a `default` part).


src/axi_test.sv
```diff
@@ -968,6 +971,12 @@ package axi_test;
           r_flight_cnt[beat.ax_id] != 0 || w_flight_cnt[beat.ax_id] !=0
         )) return 1'b0;
       end
+      if (UNIQUE_IDS) begin
+        // This master may only emit transactions with an ID that is unique among all in-flight
+        // transactions in the same direction.
+        if (is_read && r_flight_cnt[beat.ax_id] != 0) return 1'b0;
+        if (!is_read && w_flight_cnt[beat.ax_id] != 0) return 1'b0;
+      end
```
WRoenninger, line 979: I think this will only check the case where exactly one transaction per ID is in flight at any given time. This does not cover the described functionality that multiple transactions can be in flight with the same ID in the same direction.

WRoenninger (commented):
I think the changes in the RTL do cover the described functionality and LGTM.

The change to `axi_rand_master` does not cover multiple in flight transactions with the same ID to the same destination.

src/axi_test.sv
accuminium: Thanks for your review, @WRoenninger.

> I think this will only check the case where exactly one transaction per ID is in flight at any given time.

With the `UNIQUE_IDS` parameter, `axi_rand_master` fulfills the "each transaction has an ID that is unique among all in-flight transactions in the same direction" guarantee.

> This does not cover the described functionality that multiple transactions can be in flight with the same ID in the same direction.

Do you mean the "for any ID, all transactions with that ID target the same master port as all other transactions with the same ID and direction" guarantee?  If so, I agree: that guarantee is not covered.  The reason is that this guarantee is not defined for every master module but only for `axi_demux`.  `axi_rand_master` has only one master port and has no knowledge of any `axi_demux`es downstream of it, so how could it ever fulfill that guarantee?

WRoenninger: Yes I mean that guarantee. Currently the random master does not keep track of where the transactions are going. Perhaps it would be possible to tie the memory region functionality somehow into this? 

However still a tricky edge case, as then the memory regions would need to match the demux address decoding downstream. And as you mentioned the rand master has no knowledge of the demux topology downstream.

I think the guarantee with the multiple outstanding transactions should be correct as long as the master waits for the last response before switching to another slave. 

Potential issue I see is that when enabling this in modules which use the demux for error responses (like `axi_deburst` or I think the DataWidth converters), then it will lead to wrong response ordering occasionally. 
I actually had recently the same issue with this usecase of the demux that this PR is trying to solve .
In the `axi_to_axi_lite` converter the demux counters inside of `axi_deburst` were leading to unacceptable timing/area (ID width 8).

accuminium: > I think the guarantee with the multiple outstanding transactions should be correct as long as the master waits for the last response before switching to another slave.

So you would suggest to extend

"for any ID, all transactions with that ID target the same master port as all other transactions with the same ID and direction" guarantee"

to

"for any ID, all transactions with that ID target the same master port as all other *in-flight* transactions with the same ID and direction" guarantee"

?

> Potential issue I see is that when enabling this in modules which use the demux for error responses (like axi_deburst or I think the DataWidth converters), then it will lead to wrong response ordering occasionally.

Yes, it would be forbidden to enable `UniqueIds` in that case.  I think the current rules already clearly forbid this case, don't you think?

> I actually had recently the same issue with this usecase of the demux that this PR is trying to solve .
In the axi_to_axi_lite converter the demux counters inside of axi_deburst were leading to unacceptable timing/area (ID width 8).

I don't think the `UniqueIds` feature should be applied to this use case.  How about reducing the AXI ID width with one of our ID width converters before converting to AXI-Lite?

WRoenninger: I would go with the _in-flight_ wording.

> I don't think the UniqueIds feature tries to solve this use case. How about reducing the AXI ID width with one of our ID width converters before converting to AXI-Lite?

That would be a possibility, however they are not merged into master (see #103) yet.
I did a quick local `v2` using `axi_serialize` and two counters for bursts as my usecase has no interleaving accesses.

accuminium: > That would be a possibility, however they are not merged into master (see #103) yet.

True, that PR still lacks unit-level TBs -- contributions are welcome. :-)

> I would go with the *in-flight* wording.

Great, I will add that.

thommythomaso:
RTL changes look good to me. Omitting the `id_counters` should increase the scalability by a lot.

thommythomaso (approved):

WRoenninger (approved):
LGTM!

accuminium:
Thanks for your reviews!

PR #182, opened 2021-09-03
axi_lite_cdc_src_intf: Fix interchanged ports

SamuelRiedel:
It seems the `_i/_o` suffixes got interchanged when instantiating the `axi_cdc_src` module. I do not use this module directly, but Verilator gave me a warning when compiling the `axi` repo.


accuminium (approved):
LGTM, thanks!

PR #185, opened 2021-10-08
Fix latch inference on the data width converters with the Mentor Precision synthesis tool

### LINKED ISSUE #184: axi_dw_up/downsizer: Latch inference when synthesizing with Mentor Precision
When synthesizing with _Mentor Precision_, the tool infers latches in the `axi_dw_upsizer` and `axi_dw_downsizer`.
Seems the tool has an issue with the automatic variable `r_data` and `w_data` in both modules. 
Links to the of the offending codelines of `axi_dw_upsizer` and `axi_dw_downsizer`:
https://github.com/pulp-platform/axi/blob/f847575bab7269f88f6caa36f01275de2007378a/src/axi_dw_upsizer.sv#L478-L492
https://github.com/pulp-platform/axi/blob/f847575bab7269f88f6caa36f01275de2007378a/src/axi_dw_upsizer.sv#L595-L607
https://github.com/pulp-platform/axi/blob/f847575bab7269f88f6caa36f01275de2007378a/src/axi_dw_downsizer.sv#L550-L562
https://github.com/pulp-platform/axi/blob/f847575bab7269f88f6caa36f01275de2007378a/src/axi_dw_downsizer.sv#L722-L737

At least in `axi_dw_upsizer` the assignment of `r_data` should still be the default `'0`. So `r_data` is only _saving_ the initialization value, which is then conditionally overwritten.

Then however when synthesizing with _SYNOPSYS_ the latches are not there.

When adding an `else` statement to the `if` then the latch goes away at least in the first code snippet. So an idea for a fix from my side is to add the `else` condition and remove the automatic variable.

suehtamacv:
Closes #184.

This PR reworks a construct that would case latches to be infered on both the `axi_dw_downsizer` and the `axi_dw_upsizer`, when synthesizing them with the Mentor Precision.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 ### Fixed
+- `axi_dw_downsizer` and `axi_dw_upsizer` (part of `axi_dw_converter`): Avoid latch inference on the
+  Mentor Precision synthesis tool.
 - `axi_lite_cdc_src_intf`: Fix `_i` and `_o` suffixes in instantiation of `axi_cdc_src`.
 - `axi_test::axi_scoreboard`: Add default value to parameters to improve compatibility with some
   tools.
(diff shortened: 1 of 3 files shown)


accuminium (approved):
LGTM, thanks!

PR #197, opened 2022-01-19
Enforce more consistent type naming

micprog:
Unify type naming across modules, using `axi_req_t` for AXI request and `axi_lite_req_t` for axi lite request. May fix some issues for vivado (see HERO). Similar for `resp` types.


andreaskurth (changes requested):
I see the need for this change: especially the `resp_t` type collides with types of the same name in many tools. (This is a bug in the respective tool; if the tool implemented type resolution correctly according to the System Verilog standard, there would be no collision. However, as we cannot expect all tool vendors to fix this bug in the foreseeable future, we have to do our best to not trigger it.)

However, I have two concerns:
1. The types should be named according to the scheme we discussed in #153.
2. This change will break many usages of our modules. We therefore need to be careful about its rollout. We probably want to do one release where we make all the breaking changes agreed upon in #153, so that users only have to adapt their code once. However, this will require more work.

What do you think?

accuminium:
> 

Yes please!

PR #201, opened 2022-01-26
Add flat ports variant of axi_iw_converter

micprog:


accuminium:
Is there a need for a flat port variant? Put differently, which use case can work neither with the struct nor the interface variant?

micprog:
> Is there a need for a flat port variant? Put differently, which use case can work neither with the struct nor the interface variant?

I am not very familiar with the differences between verilog and systemverilog, however the flat variant is used in a top-level fpga verilog file (or template) for HERO: https://github.com/pulp-platform/hero/blob/ff8773fbd347b79d9fdd3820e11cf3ec3b9d8b6d/hardware/fpga/src/pulp.template_v#L192
If this can be replaced with an interface variant, then the flat variant is not needed.

accuminium:
> I am not very familiar with the differences between verilog and systemverilog, however the flat variant is [used in a top-level fpga verilog file (or template) for HERO](https://github.com/pulp-platform/hero/blob/ff8773fbd347b79d9fdd3820e11cf3ec3b9d8b6d/hardware/fpga/src/pulp.template_v#L192). If this can be replaced with an interface variant, then the flat variant is not needed.

I see, flat port lists are required to instantiate the module from Verilog code. But as far as I know, Vivado is the only tool that forces us to still use Verilog code (and only for top-level code of modules). So I think this is such a special case that we do not want to mainline and maintain the code in this repository.

I thus suggest to keep that `_flat` module in the downstream repository.

micprog:
Makes sense, I will keep the change only in the downstream repository.

PR #208, opened 2022-02-18
axi_xbar: Work around constant function calls for VCS

bluewww:
VCS seems to be unable to handle constant function calls whose argument is a `struct` passed into a module as `parameter`.  `axi_xbar` became prone to this problem with commit d24303f0df9.  This commit works around the problem by calling a system function instead of a constant function.

### DIFF
--- a/src/axi_xbar.sv
+++ b/src/axi_xbar.sv
@@ -34,6 +34,10 @@ import cf_math_pkg::idx_width;
   parameter type mst_req_t          = logic,
   parameter type mst_resp_t         = logic,
   parameter type rule_t             = axi_pkg::xbar_rule_64_t
+`ifdef VCS
+  , localparam int unsigned MstPortsIdxWidth =
+      (Cfg.NoMstPorts == 32'd1) ? 32'd1 : unsigned'($clog2(Cfg.NoMstPorts))
+`endif
 ) (
   input  logic                                                          clk_i,
   input  logic                                                          rst_ni,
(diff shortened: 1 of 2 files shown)


accuminium (changes requested):
Thanks for your efforts to debug this and propose a first solution, @bluewww!

I agree that we need to make a reasonable effort for finding workarounds to VCS limitations, and that might introduce more `ifdef`s. However, as `cf_math_pkg::idx_width()` is not equivalent to `$clog2()` when the argument has value `1`, additional work is required to solve those cases. One option could be to introduce a VCS-only dependent parameter:
```sv
`ifdef VCS
parameter int unsigned MstPortsIdxWidth = Cfg.NoMstPorts == 32'd1 ? 32'd1 : unsigned'($clog2(Cfg.NoMstPorts))
`endif
```

bluewww:
I caught this problem by issuing an elaboration time error see https://github.com/pulp-platform/axi/pull/208/commits/59b2f88e4fad0aea9aa72a4425497160431427b7#diff-e525faff2463489e38729f200de6530d6911b2d11b58fc46928d1e1e884683c9R56-R60

bluewww:
I rebased and dropped the fixes which you address in this PR https://github.com/pulp-platform/axi/pull/210

As you suggested I now introduce vcs dependent parameters. This works too and for all cases of `Cfg.NoMstPorts >=1` to work

The placement of the commas is quite unsatisfying maybe you have a better idea.

src/axi_xbar.sv
```diff
@@ -34,6 +34,10 @@ import cf_math_pkg::idx_width;
   parameter type mst_req_t          = logic,
   parameter type mst_resp_t         = logic,
   parameter type rule_t             = axi_pkg::xbar_rule_64_t
+`ifdef VCS
+  , localparam int unsigned MstPortsIdxWidth    = Cfg.NoMstPorts == 32'd1 ? 32'd1 : unsigned'($clog2(Cfg.NoMstPorts))
+  , localparam int unsigned MstPortsIdxWidthOne = Cfg.NoMstPorts == 32'd1 ? 32'd1 : unsigned'($clog2(Cfg.NoMstPorts + 1))
```
accuminium, line 39: This `localparam` is not used in the port list, so please move it into the module body (e.g., just before line 61).

accuminium (changes requested):

accuminium (approved):
LGTM, thanks!

PR #212, opened 2022-02-25
Add `axi_dumper` and `axi_dumper_interpret` script.

micprog:
This module dumps a log from an AXI bus, which can be analyzed with the interpret script for debugging purposes.

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -71,6 +71,7 @@ sources:
 
   - target: simulation
     files:
+      - src/axi_dumper.sv
       - src/axi_sim_mem.sv
       - src/axi_test.sv
 
--- a/scripts/axi_dumper_interpret.py
+++ b/scripts/axi_dumper_interpret.py
@@ -1,0 +1,213 @@
+# Copyright (c) 2020 ETH Zurich, University of Bologna
+# All rights reserved.
+#
+# Copyright and related rights are licensed under the Solderpad Hardware
+# License, Version 0.51 (the "License"); you may not use this file except in
+# compliance with the License.  You may obtain a copy of the License at
+# http://solderpad.org/licenses/SHL-0.51. Unless required by applicable law
+# or agreed to in writing, software, hardware and materials distributed under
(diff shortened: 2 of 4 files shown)


micprog:
I recently came across this which might implement similar functionality: https://github.com/pulp-platform/axi/blob/db457aa8d8996d87ca72486a9b15873042e04d38/src/axi_test.sv#L2195

thommythomaso:
> I recently came across this which might implement similar functionality:
> 
> https://github.com/pulp-platform/axi/blob/db457aa8d8996d87ca72486a9b15873042e04d38/src/axi_test.sv#L2195

I like the idea of the `axi_dumper` overall better. The output is machine-parsable but still human-understandable making this module much more helpful than the classical logger which only creates human-readable output.

scripts/axi_dumper_interpret.py
```diff
@@ -0,0 +1,264 @@
...
+#
+# Copyright and related rights are licensed under the Solderpad Hardware
+# License, Version 0.51 (the "License"); you may not use this file except in
+# compliance with the License.  You may obtain a copy of the License at
+# http://solderpad.org/licenses/SHL-0.51. Unless required by applicable law
+# or agreed to in writing, software, hardware and materials distributed under
+# this License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+
```
thommythomaso, line 12: Maybe add your name here

Bender.yml
```diff
@@ -71,6 +71,7 @@ sources:
 
   - target: simulation
     files:
+      - src/axi_dumper.sv
```
thommythomaso, line 74: We will have to update the other two dependency tools as well. I can do that in the release commit though.

thommythomaso (commented):
LGTM. I will do another pass once you feel it's done.

thommythomaso (approved):
:+1:

PR #216, opened 2022-03-09
axi_sim_mem: Keep R response stable while not accepted

SamuelRiedel:
If the R response was not accepted right away, a new write could change the read data violating the stability guarantee of AXI.


andreaskurth (approved):
LGTM. Thanks for finding and fixing this bug, @SamuelRiedel!

PR #217, opened 2022-03-09
axi_sim_mem: Add monitoring interface

SamuelRiedel:
Add a monitoring interface into the `axi_sim_mem` to observe the Point of Coherency between the write and the read channel.

#216 Has to be merged first.

### DIFF
--- a/src/axi_sim_mem.sv
+++ b/src/axi_sim_mem.sv
@@ -46,7 +46,39 @@ module axi_sim_mem #(
   /// AXI4 request struct
   input  axi_req_t axi_req_i,
   /// AXI4 response struct
-  output axi_rsp_t axi_rsp_o
+  output axi_rsp_t axi_rsp_o,
+  /// Memory monitor write valid.  All `mon_w_*` outputs are only valid if this signal is high.
+  /// A write to the memory is visible on the `mon_w_*` outputs in the clock cycle after it has
+  /// happened.
+  output logic                 mon_w_valid_o,
+  /// Memory monitor write address
+  output logic [AddrWidth-1:0] mon_w_addr_o,
+  /// Memory monitor write data
+  output logic [DataWidth-1:0] mon_w_data_o,
+  /// Memory monitor write ID
+  output logic [IdWidth-1:0]   mon_w_id_o,
+  /// Memory monitor write user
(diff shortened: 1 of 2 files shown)


src/axi_sim_mem.sv
```diff
@@ -176,23 +227,69 @@ module axi_sim_mem #(
...
+      mon_w_user_o <= #(ApplDelay) mon_w.user;
+      mon_w_beat_count_o <= #(ApplDelay) mon_w.beat_count;
+      mon_w_last_o <= #(ApplDelay) mon_w.last;
+      mon_r_valid_o <= #(ApplDelay) mon_r.valid;
+      mon_r_addr_o <= #(ApplDelay) mon_r.addr;
+      mon_r_data_o <= #(ApplDelay) mon_r.data;
+      mon_r_id_o <= #(ApplDelay) mon_r.id;
+      mon_r_user_o <= #(ApplDelay) mon_r.user;
+      mon_r_beat_count_o <= #(ApplDelay) mon_r.beat_count;
+      mon_r_last_o <= #(ApplDelay) mon_r.last;
```
andreaskurth, line 289: What is the rationale behind waiting one cycle for the monitor port, i.e., to expose read and writes that happened in the *previous* cycle at the monitor port?

SamuelRiedel: The problems are the `ApplDelay` and `AcqDelay`. We only know whether we are writing until after the `AcqDelay`:

https://github.com/pulp-platform/axi/blob/0a839bf2e31500c8d708971f20fc84de86000add/src/axi_sim_mem.sv#L135-L136

This means that we could only provide the monitoring output of the `w_channel` from the `AcqDelay` onwards in the same cycle, which breaks the whole application/acquisition delay construct. Unfortunately, I don't see a way to provide the monitor output at `ApplDelay` if we want to only sample the inputs after `AcqDelay`, so I added a register stage.

andreaskurth: I see, thanks for the explanation.

I have added a `fixup!` commit to describe this at the ports and in the implementation. If you agree, please apply the fixup. Once this is resolved, we can go ahead with merging.

SamuelRiedel: Perfect, thanks :+1:

PR #221, opened 2022-03-16
axi_sim_mem: Propagate the ar_user signal correctly to the monitor

SamuelRiedel:
The AR channel's user signal was not correctly propagated to the monitor interface since we do not reflect the user signal. As the R and AR channels do not have the same user signals, it makes sense to use the AR's signal for the monitoring interface since this is the one the user has control over and might want to observe.


andreaskurth (approved):
I agree, thanks!

PR #225, opened 2022-03-30
axi_{,lite_}{,de}mux: Add missing spill registers for single master/slave port configurations

### LINKED ISSUE #223: `axi_(lite_)_(de)mux`: Single master/slave edge case ignores spill parameters
In each of the modules:

* `axi_mux`
* `axi_demux`
* `axi_lite_mux`
* `axi_lite_demux`

There is a generate case in which the modules become feedthroughs. For example:  https://github.com/pulp-platform/axi/blob/e17b99a7ebaef3c6d908a739b47a2acb324539a7/src/axi_mux.sv#L70-L73

Unfortunately, these cases ignore the `Spill*` parameters. This can lead to unintended input-to-output paths in such modules and in particular in `1xN` and `Nx1` crossbars built on them, which can cause serious synthesis issues and timing loops.

A correct design would instantiate `spill_register`s for each channel with the `Bypass` parameter set by the `Spill*` parameter for each channel. Better yet, `axi_cut` could be adapted to enable the cutting of individual channels and inserted in these generate cases.

andreaskurth:
Fixes #223.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 ### Fixed
+- `axi_demux`: Add missing spill registers for configurations with a single master port.
 - `axi_demux_intf`: Add missing parameter (`ATOP_SUPPORT`) to optionally disable support for atomic
   operations.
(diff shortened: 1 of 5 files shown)


paulsc96 (approved):
Looks good to me :+1:.

Issue #191, opened 2021-12-08
Add AXI FIFO


andreaskurth:
We have [`axi_cut`](https://github.com/pulp-platform/axi/blob/master/src/axi_cut.sv) and [`axi_multicut`](https://github.com/pulp-platform/axi/blob/master/src/axi_multicut.sv) for pipelining ([to be unified](https://github.com/pulp-platform/axi/issues/154)), but we currently don't have a module for buffering AXI beats in FIFOs. This gap can be filled by a new `axi_fifo` module.

Implementing this module should be as simple as instantiating one FIFO per channel and make each configurable in terms of depth and fall-through property.

Testing this module should be as simple as monitoring each slave/master port pair of channels and compare their beats with one dynamic queue per channel.

thommythomaso:
We will merge PR #230 now with a shared depth for all channels. Should the need arise for more fine-granular control over the FIFO configuration this can be done in a future PR.

PR #230, opened 2022-04-28
Add module `axi_fifo`

noah95:
This adds the `axi_fifo` module that puts a `fifo_v3` in all five AXI4 channels to buffer requests and responses.

Solves #191.

## ToDo
- [ ] Parametrization OK or should it be split into Depths/FallThrough per channel?
- [ ] Is the verification good enough?


micprog (approved):
LGTM :+1:

PR #231, opened 2022-05-02
axi_demux: Eliminate unnecessary AW channel stalls

micprog:
Eliminate unnecessary stalls of AW channel when the AR channel has
reached its maximum number of transactions.  Prior to this fix,
`axi_demux` would always stall AWs while read transactions were at their
maximum (that is, while `MaxTrans` read transactions were outstanding).
However, this stall is only required when the AW that is being handled
by `axi_demux` is an atomic operation (ATOP) that entails an R response.
This fix therefore removes unnecessary stalls as well as an unnecessary
dependency between reads and writes.  The integrity of data or
transactions was not affected by this problem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 ### Fixed
+- `axi_demux`: Eliminate unnecessary stalls of AW channel when the AR channel has reached its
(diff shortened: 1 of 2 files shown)


andreaskurth (approved):
Thanks for this fix, @micprog!

I agree with the change and I think it is an important one, so I have extended the Changelog entry, to make sure users understand it when it shows up in the release notes. Please let me know what you think.

micprog:
LGTM, thanks for making things clearer!

PR #242, opened 2022-08-04
axi_sim_mem: Add the possibility to emit read and write errors

thommythomaso:


src/axi_sim_mem.sv
```diff
@@ -104,7 +108,12 @@ module axi_sim_mem #(
   } monitor_t;
 
   monitor_t mon_w, mon_r;
-  logic [7:0] mem[addr_t];
+  logic [7:0]     mem[addr_t];
+  axi_pkg::resp_t rerr[addr_t] = '{default: axi_pkg::BURST_FIXED};
+  axi_pkg::resp_t werr[addr_t] = '{default: axi_pkg::BURST_FIXED};
```
andreaskurth, line 113: Shouldn't these be `axi_pkg::RESP_OKAY`?

src/axi_sim_mem.sv
```diff
@@ -156,16 +165,20 @@ module axi_sim_mem #(
               if (axi_req_i.w.strb[i_byte]) begin
                 automatic addr_t byte_addr = (addr / StrbWidth) * StrbWidth + i_byte;
                 mem[byte_addr] = axi_req_i.w.data[i_byte*8+:8];
+                error_happened = error_happened | werr[byte_addr];
```
andreaskurth, line 168: Suppose one byte has `RESP_EXOKAY` and the next byte has `RESP_SLVERR`. Then this OR would result in `RESP_DECERR`, right? That seems wrong to me.

andreaskurth (changes requested):

src/axi_sim_mem.sv
thommythomaso: Yes of course. Sorry for that.

src/axi_sim_mem.sv
thommythomaso: For writes, we only get one answer once the entire burst is completed. We decided here that if one byte of a write burst throws an error, the entire burst will be completed with a b answer carrying said error.

micprog: @thommythomaso I believe the concern here is that `RESP_EXOKAY = 01` while `RESP_SLVERR = 10` which through OR would equate to `RESP_DECERR = 11`, which I agree is not what the response should be. Overall I'd argue as this module does not support ATOPs or locking, all responses should be either `RESP_OKAY = 00` or `RESP_SLVERR = 10` (and maybe `RESP_DECERR`), as `RESP_EXOKAY` does not make sense.
Additionally, following the specification:

> A7.2.5 on responses to exclusive access: An exclusive read has one or more response beats. These can be a mixture of EXOKAY, SLVERR and DECERR or a mixture of OKAY, SLVERR and DECERR. Mixing EXOKAY and OKAY responses in the same transaction is not permitted.
> B1.3.1 on conversion rules: Where a wide transaction is converted to multiple narrower transactions, the component responsible for the conversion must combine the responses to all of the narrower transactions, to produce a single response for the original transaction. Any error response is sticky. If both a SLVERR and a DECERR are received then the first response received is used for the combined response.

I would therefore argue that this module may need to limit possible responses to only `RESP_OKAY` and `RESP_SLVERR`, possibly also `RESP_DECERR` if needed.

Does something like this make sense for you? Something similar is also needed for the `rerr`.
```suggestion
                error_happened = (error_happened != axi_pkg::RESP_OKAY) ? error_happened : ((werr[byte_addr] == axi_pkg::RESP_EXOKAY) ? axi_pkg::RESP_OKAY : werr[byte_addr]);

```

thommythomaso: I get this point. However, errors have to be deliberately introduced into the error arrays by the user (it's not possible to inject them from the AXI bus) so from my point of view, a mix between `SLVERR` and `EXOKAY` should be counted as a user error. 
I agree that is better to be resilient against such user errors so this line should be changed. Should https://github.com/pulp-platform/axi/blob/db457aa8d8996d87ca72486a9b15873042e04d38/src/axi_pkg.sv#L264-L291 be used to merge the responses here?

micprog: `resp_precedence` looks good to me, I was unaware of this feature.

thommythomaso: I will rewrite the response accumulation using `resp_precedence`.

micprog (approved):
LGTM, thanks for the fix

andreaskurth (approved):
Thanks for the discussion and implementing the feedback!

PR #213, opened 2022-02-25
AXI4 on chip memory slave

micprog:
Rebase of the relevant changes in #115 on the current master

This adds two modules:
- `axi_to_mem`: Slave module, max throughout simultaneous read/writes 50%, read or write 100%.
- `axi_to_mem_banked`: With enough banks 100% throughput with simultaneous reads/writes.

Notes to Bender:
For the dependencies pulp-platform/tech_cells_generic has been directly added. In there the tc_sram module is used for the testbench to serve a memory model.

Open tasks
- [x] Rework `mem_to_banks` to ensure no writes are issued with `strb='0`. While not illegal, this can cause issues (e.g. when converting back to APB3) and unnecessary contention on an interconnect
- [x] Add an interleaved variant to allow reads and writes to bypass each other
- [x] Add a split variant to completely separate the read and write channel to individual mem ports when connecting to an interconnect

PR #115, opened 2020-06-25
AXI4+ATOP on chip memory slave

WRoenninger:
This adds two modules:
- `axi_to_mem`: Salve module, max throughout simultaneous read/writes 50%, read or write 100%.
- `axi_to_mem_banked`: With enough banks 100% throughput with simultaneous reads/writes.

Notes to Bender:
For the  dependencies `pulp-platform/tech_cells_generic` has been directly added. In there the `tc_sram` module is used for the testbench to serve a memory model.
-> For all conflicts in the dependency tree of Bender, the versions specified by the Bender.yml of this repo has to be used.

Edit: Now bender.lock is also checked in so CI should pass now

This depends on changes form #116.

PR #244, opened 2022-08-05
AXI4 on chip memory slave

micprog:
This PR adds the `axi_to_mem` module to translate AXI4 to a simple memory protocol. This PR partially replaces #115 and #213.

### DIFF
--- a/src/axi_to_mem.sv
+++ b/src/axi_to_mem.sv
@@ -1,0 +1,697 @@
+// Copyright 2020 ETH Zurich and University of Bologna.
+// Copyright and related rights are licensed under the Solderpad Hardware
+// License, Version 0.51 (the "License"); you may not use this file except in
+// compliance with the License. You may obtain a copy of the License at
+// http://solderpad.org/licenses/SHL-0.51. Unless required by applicable law
+// or agreed to in writing, software, hardware and materials distributed under
+// this License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+// Authors:
+// - Andreas Kurth <akurth@iis.ee.ethz.ch>
+// - Wolfgang Roenninger <wroennin@iis.ee.ethz.ch>
+
+`include "common_cells/registers.svh"
+/// AXI4+ATOP slave module which translates AXI bursts into a memory stream.
+/// If both read and write channels of the AXI4+ATOP are active, both will have an
+/// utilization of 50%.
+module axi_to_mem #(
+  /// AXI4+ATOP request type. See `include/axi/typedef.svh`.
+  parameter type         axi_req_t  = logic,
+  /// AXI4+ATOP response type. See `include/axi/typedef.svh`.
+  parameter type         axi_resp_t = logic,
+  /// Address width, has to be less or equal than the width off the AXI address field.
+  /// Determines the width of `mem_addr_o`. Has to be wide enough to emit the memory region
+  /// which should be accessible.
+  parameter int unsigned AddrWidth  = 0,
+  /// AXI4+ATOP data width.
+  parameter int unsigned DataWidth  = 0,
+  /// AXI4+ATOP ID width.
+  parameter int unsigned IdWidth    = 0,
+  /// Number of banks at output, must evenly divide `DataWidth`.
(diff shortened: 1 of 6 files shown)


src/axi_to_mem.sv
```diff
@@ -0,0 +1,697 @@
...
+    .mem_atop_o,
+    .mem_we_o,
+    .mem_rvalid_i,
+    .mem_rdata_i
+  );
+endmodule
+
+/// Split memory access over multiple parallel banks, where each bank has its own req/gnt
+/// request and valid response direction.
+module mem_to_banks #(
```
thommythomaso, line 558: Should this module be part of common cells?

src/axi_to_mem.sv
```diff
@@ -0,0 +1,697 @@
...
+
+// Authors:
+// - Andreas Kurth <akurth@iis.ee.ethz.ch>
+// - Wolfgang Roenninger <wroennin@iis.ee.ethz.ch>
+
+`include "common_cells/registers.svh"
+/// AXI4+ATOP slave module which translates AXI bursts into a memory stream.
+/// If both read and write channels of the AXI4+ATOP are active, both will have an
+/// utilization of 50%.
+module axi_to_mem #(
```
thommythomaso, line 19: Does it make sense to directly include an enhanced version of this module than can accept at least one read and write request on the AXI subordinate interface? With the blocking of the read channel while writes happen we create an implicit coupling between the channels, which whilst being allowed for subordinate modules, could impede the flow in the AXI system.

src/axi_to_mem.sv
micprog: I currently don't see another use-case, but am open to moving this to [common_cells](https://github.com/pulp-platform/common_cells). I took this from @WRoenninger's original PR #115 and updated it with an optional feature that caused issues for me.

src/axi_to_mem.sv
micprog: In accordance with @andreaskurth's request in #213 I split up the PR into separate ones, this is the first with `axi_to_mem`. The next will include `axi_to_mem_banked` and tests of that module, which will also test `axi_to_mem`. I believe you mean the `axi_to_mem_interleaved` which will be a separate request. For access to a memory protocol interconnect I will add a fourth PR with `axi_to_mem_split`. I agree with @andreaskurth, the best approach is probably to separate these modules to individual PRs for review, however I would encourage to add them all in a single version bump.

thommythomaso: I understand.

micprog:
> We do not have a testbench for this module, right?

#248 adds a banked variant using `axi_to_mem`, as well as a testbench for the banked variant, consequently also testing `axi_to_mem`.

thommythomaso (approved):
LGTM!

src/axi_to_mem.sv
```diff
@@ -648,19 +661,52 @@ module mem_to_banks #(
...
   end
 
   // Grant output if all our requests have been granted.
-  assign gnt_o = (&req_ready) & (&resp_ready);
+  assign gnt_o = (&req_ready) & (&resp_ready) & !dead_write_fifo_full;
+
+  if (HideStrb) begin : gen_dead_write_fifo
+    fifo_v3 #(
+      .FALL_THROUGH ( 1'b1     ),
+      .DEPTH        ( MaxTrans+1 ),
```
andreaskurth, line 691: Why +1?

andreaskurth (approved):
LGTM, thanks for the effort for getting this (finally) merged! :slightly_smiling_face:

src/axi_to_mem.sv
micprog: When adding this option, +1 was necessary to avoid unnecessary stalls due to flow control, as the FIFO will signal it is full even if an element is being popped, thus not allowing another to be pushed. +1 is a quick fix for this, as modifying the underlying fifo may introduce combinatorial loops. Simplest example is for single cycle latency memory with `MaxTrans=1`, where consecutive writes would experience a stall.

thommythomaso: If the FIFO is full it signals it being not ready to accept data upstream. A „full push“ theoretically would be allowed if its getting popped in the same cycle, but afaik the push logic does not know this (otherwise it would not be decoupled).

Maybe a FIFO coupled to a credit counter would solve this?

thommythomaso: But lets keep the `+1` for now.

PR #248, opened 2022-08-09
AXI4 banked on chip memory slave

micprog:
This PR adds a banked variant of `axi_to_mem` for higher throughput. This PR also adds a test for `axi_to_mem_banked`, which also tests `axi_to_mem` as it is called within the new module. This PR partially replaces #115 and #213 and depends on #244.

### DIFF
--- a/src/axi_to_mem_banked.sv
+++ b/src/axi_to_mem_banked.sv
@@ -107,7 +107,6 @@ module axi_to_mem_banked #(
     WriteAccess = 1'b1
   } access_type_e;
   typedef logic [AxiAddrWidth-1:0] axi_addr_t;
-  // typedef logic [TcdmDataWidth-1:0] tcdm_data_t;
 
   /// Payload definition which is sent over the xbar between the macros and the read/write unit.
   typedef struct packed {
@@ -146,7 +145,7 @@ module axi_to_mem_banked #(
     .axi_req_t   ( axi_req_t     ),
     .axi_resp_t  ( axi_resp_t    ),
     .NoMstPorts  ( 32'd2         ),
-    .MaxTrans    ( 32'd4         ), // allow multiple Ax vectors to not starve W channel
+    .MaxTrans    ( MemLatency+2  ), // allow multiple Ax vectors to not starve W channel
     .AxiLookBits ( 32'd1         ), // select is fixed, do not need it
     .UniqueIds   ( 1'b0          ),
     .FallThrough ( 1'b1          ),
--- a/src/axi_to_mem_banked.sv
+++ b/src/axi_to_mem_banked.sv
@@ -1,0 +1,430 @@
+// Copyright 2020 ETH Zurich and University of Bologna.
+// Copyright and related rights are licensed under the Solderpad Hardware
+// License, Version 0.51 (the "License"); you may not use this file except in
+// compliance with the License. You may obtain a copy of the License at
+// http://solderpad.org/licenses/SHL-0.51. Unless required by applicable law
+// or agreed to in writing, software, hardware and materials distributed under
+// this License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+// Authors:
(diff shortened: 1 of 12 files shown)


src/axi_to_mem_banked.sv
```diff
@@ -0,0 +1,428 @@
...
+  localparam int unsigned BankSelWidth  = cf_math_pkg::idx_width(MemNumBanks);
+  typedef logic [BankSelWidth-1:0] xbar_sel_t;
+
+  // Typedef for defining the channels
+  typedef enum logic {
+    ReadAccess  = 1'b0,
+    WriteAccess = 1'b1
+  } access_type_e;
+  typedef logic [AxiAddrWidth-1:0] axi_addr_t;
+  // typedef logic [TcdmDataWidth-1:0] tcdm_data_t;
```
thommythomaso, line 110: please remove commented code if it is no longer used.

src/axi_to_mem_banked.sv
```diff
@@ -0,0 +1,428 @@
...
+    /// Selection signal for response routing
+    xbar_sel_t sel;
+    /// Selection is valid
+    logic      valid;
+  } read_sel_t;
+
+  axi_req_t  [1:0] mem_axi_reqs;
+  axi_resp_t [1:0] mem_axi_resps;
+
+  // Fixed select `axi_demux` to split reads and writes to the two `axi_to_mem`
```
thommythomaso, line 137: Do we really need a demux here to split reads and writes? The channels should be independent so a simple passive splitter should suffice, right?

src/axi_to_mem_banked.sv
```diff
@@ -0,0 +1,428 @@
...
+    .AxiIdWidth  ( AxiIdWidth    ),
+    .aw_chan_t   ( axi_aw_chan_t ),
+    .w_chan_t    ( axi_w_chan_t  ),
+    .b_chan_t    ( axi_b_chan_t  ),
+    .ar_chan_t   ( axi_ar_chan_t ),
+    .r_chan_t    ( axi_r_chan_t  ),
+    .req_t       ( axi_req_t     ),
+    .resp_t      ( axi_resp_t    ),
+    .NoMstPorts  ( 32'd2         ),
+    .MaxTrans    ( 32'd4         ), // allow multiple Ax vectors to not starve W channel
```
thommythomaso, line 148: This should be a top-level parameter -> in some systems final memories might have more than a few cycles latency...

src/axi_to_mem_banked.sv
```diff
@@ -0,0 +1,428 @@
...
+    .AxiLookBits ( 32'd1         ), // select is fixed, do not need it
+    .FallThrough ( 1'b1          ),
+    .SpillAw     ( 1'b1          ),
```
thommythomaso, line 151: A lot of hardware; cutting all channels unconditionally.

src/axi_to_mem_banked.sv
```diff
@@ -0,0 +1,428 @@
...
+        .d_i    ( r_shift_inp ),
+        .d_o    ( r_shift_oup )
+      );
+    end
+  end
+
+  // Xbar to arbitrate data over the different memory banks
+  xbar_payload_t [MemNumBanks-1:0] mem_payload;
+
+  stream_xbar #(
```
thommythomaso, line 254: Wouldn't a TCDM interconnect be the wiser choice here?

thommythomaso (commented):
Looks in general good, I have some comments.

src/axi_to_mem_banked.sv
micprog: Setting to `MemLatency + 2` to ensure latency and cuts are filled

src/axi_to_mem_banked.sv
micprog: Maybe, but that should instantiate the same common_cells module, and this way fewer dependencies for the AXI repository are needed.

src/axi_to_mem_banked.sv
micprog: You're right, we should fix this in a future PR

src/axi_to_mem_banked.sv
micprog: Haven't thought it through completely, possibly future PR

PR #250, opened 2022-08-17
First batch of fixes and improvements originating from the `axi_llc` branch

thommythomaso:
Included are the following changes:
- axi_test:
  - axi_test: Use `beat_addr` function
  - axi_test: Add mapped mode to random classes
  - axi_test: Additions to scoreboard class
- axi_lite_demux: 
  - axi_lite_demux: Increase VSIM compatibility
- axi_demux (***amended***): 
  - axi_demux: Replace w_fifo with a write counter

### DIFF
--- a/src/axi_lite_demux.sv
+++ b/src/axi_lite_demux.sv
@@ -426,8 +426,15 @@ module axi_lite_demux #(
     );
 
     // connect the response if the FIFO has valid data in it
-    assign slv_r_chan      = (!r_fifo_empty) ? mst_resps_i[r_select].r : '0;
-    assign slv_r_valid     =  ~r_fifo_empty  & mst_resps_i[r_select].r_valid;
+    always_comb begin
+       slv_r_chan  = '0;
+       slv_r_valid = '0;
+       if (!r_fifo_empty) begin
+          slv_r_chan  = mst_resps_i[r_select].r;
+          slv_r_valid = mst_resps_i[r_select].r_valid;
+       end
+    end
+
     for (genvar i = 0; i < NoMstPorts; i++) begin : gen_mst_r
       assign mst_reqs_o[i].r_ready = ~r_fifo_empty & slv_r_ready & (r_select == select_t'(i));
     end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 - `axi_to_mem`: AXI4+ATOP slave to control on chip memory.
+- `axi_test`: Add `mapped` mode to the random classes as well as additional functionality to the
+  scoreboard class.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - `axi_lite_mux`: Reduce complexity of W channel at master port by removing an unnecessary
   multiplexer.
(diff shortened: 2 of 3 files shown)


thommythomaso:
I can divide the PR if a per-file workflow is requested.

src/axi_demux.sv
```diff
@@ -275,19 +276,23 @@ module axi_demux #(
...
+          /// There is a valid AW vector make the id lookup and go further, if it passes.
+          /// Also stall if previous transmitted AWs still have active W's in flight.
+          /// This prevents deadlocking of the W channel. The counters are there for the
+          /// Handling of the B responses.
+          if (slv_aw_valid &&
+                ((w_open == '0) || (w_select == slv_aw_chan_select.aw_select)) &&
+                (!aw_select_occupied || (slv_aw_chan_select.aw_select == lookup_aw_select))) begin
             // connect the handshake
             aw_valid     = 1'b1;
             // push arbitration to the W FIFO regardless, do not wait for the AW transaction
```
micprog, line 294: Might be worth changing this comment

src/axi_demux.sv
```diff
@@ -331,32 +336,36 @@ module axi_demux #(
...
+      .down_i     ( w_cnt_down            ),
+      .d_i        ( '0                    ),
+      .q_o        ( w_open                ),
+      .overflow_o ( /*not used*/          )
     );
 
+    `FFLARN(w_select_q, slv_aw_chan_select.aw_select, w_cnt_up, select_t'(0), clk_i, rst_ni)
+    assign w_select       = (|w_open) ? w_select_q : slv_aw_chan_select.aw_select;
+    assign w_select_valid = w_cnt_up | (|w_open);
+
```
micprog, line 368: Are there performance implications with this change? The demux will no longer forward all outstanding AW beats, but these should always be accompanied by their W beats. This homogenizes the behavior with the `axi_demux_id_counters`.

micprog (commented):
In addition to the other comments, I suggest to also update the changelog.

src/axi_lite_demux.sv
```diff
@@ -426,8 +426,21 @@ module axi_lite_demux #(
...
+       if (!r_fifo_empty) begin
+          slv_r_chan  = mst_resps_i[r_select].r;
+          slv_r_valid = mst_resps_i[r_select].r_valid;
+       end
+    end
+    `else
     assign slv_r_chan      = (!r_fifo_empty) ? mst_resps_i[r_select].r : '0;
     assign slv_r_valid     =  ~r_fifo_empty  & mst_resps_i[r_select].r_valid;
+    `endif
+
```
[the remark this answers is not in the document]
luca-valente: It is both. I think we should use only the compatible one. I left it because I know we like fancy :rocket:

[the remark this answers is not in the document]
thommythomaso: I will delete the fancy option then.

src/axi_demux.sv
thommythomaso: Do we have non-functional regression tests?

thommythomaso: The change seems to be required by the `AXI LLC`.

luca-valente: @thommythomaso yes, apparently the LLC triggers a corner case stalling the `demux`. That's why I cherry-picked the commit.

thommythomaso: This change is amended, it will be handled in a future PR.

luca-valente: It is fine for me :smile: but I do not understand the rationale. If the `axi_demux` has a bug causing a deadlock, and this is the fix, why should we exclude the commit? Wouldn't it make sense to merge this commit and, later on, try to remove the extra cycle latency?

PR #253, opened 2022-08-29
`axi_demux`: Replace write FIFO (`w_fifo`) with a write credit counter

thommythomaso:


WRoenninger:
From #166:

axi_demux: Remove FIFO between AW and W channel, is now a register plus counter.
Prevents AWs to be emmitted downstream to a different master port as long as Ws
are still in flight to another. This prevents deadlock, if there is stalling
downstream.

src/axi_demux.sv
```diff
@@ -331,32 +330,36 @@ module axi_demux #(
...
+        .push_i                       ( w_cnt_up                                      ),
         .pop_axi_id_i                 ( slv_b_chan.id[0+:AxiLookBits]                 ),
         .pop_i                        ( slv_b_valid & slv_b_ready                     )
       );
       // pop from ID counter on outward transaction
     end
 
-    // FIFO to save W selection
-    fifo_v3 #(
-      .FALL_THROUGH ( FallThrough ),
```
micprog, line 343: Due to the removal of the `FallThrough` parameter here, this parameter is no longer needed for the demux module. I suggest removal, although this will introduce backwards incompatibilities.

PR #256, opened 2022-08-29
axi_to_mem split & interleaved

micprog:
Interleaved and split variant of axi_to_mem


thommythomaso (approved):
LGTM :+1: Can't wait for these modules to be finally released :D

PR #264, opened 2022-09-27
vivado xsim : add XILINX_SIMULATOR ifdefs replace XSIM ifdef

skokvermon:
replace XSIM ifdef by XILINX_SIMULATOR  as described in https://docs.xilinx.com/r/en-US/ug900-vivado-logic-simulation/Predefined-Macros
depends on https://github.com/pulp-platform/common_cells/pull/139

Some more benchs are working

However there are segfaults on a lot of them.

### DIFF
--- a/src/axi_demux.sv
+++ b/src/axi_demux.sv
@@ -625,7 +625,7 @@ module axi_demux #(
 // Validate parameters.
 // pragma translate_off
 `ifndef VERILATOR
-`ifndef XSIM
+`ifndef XILINX_SIMULATOR
     initial begin: validate_params
       no_mst_ports: assume (NoMstPorts > 0) else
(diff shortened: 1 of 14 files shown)


thommythomaso:
We have internally discussed this matter. We can change `XSIM` to `XILINX_SIMULATOR` and guard/implement alternatives to make the TBs work in xsim if:
* the changes happening globally and in the dependencies first, e.g., change ` XSIM` to `XILINX_SIMULATOR` in common_cells first
* the testbenches (or most of them) do run
* the target is for sure not used in fusesoc for a different purpose (as `XSIM`was introduced by olofk)
* the code of the testbench will not look too fragmented due to too many `ifdefs`

imchenwu:
Close as #434 is merged.

PR #269, opened 2022-09-29
Add a simple makefile to elaborate and run simulations

thommythomaso:


Makefile
```diff
@@ -0,0 +1,40 @@
...
+# or agreed to in writing, software, hardware and materials distributed under
+# this License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+#
+# Authors:
+# - Thomas Benz <tbenz@iis.ee.ethz.ch>
+
+VSIM        ?= "questa-2022.3 vsim"
+SYNOPSYS_DC ?= "synopsys-2022.03 dc_shell"
```
micprog, line 14: I'd argue we keep agnostic versions here (i.e. only `vsim` and `dc_shell`). If you prefer, we can do default agnostic, but have an automatic switch if the file `/etc/iis.version` (or folder `/usr/sepp`) is found to differentiate internal systems.

micprog (approved):
In general LGTM. I'm not 100% sure we need to have this in addition to the compile and simulate scripts, but I don't think it hurts. As noted I'd change the default tool version for compatibility. It might also make sense to change the CI to use the Makefile instead of the scripts for cleaner build, but that is up to you.

Makefile
thommythomaso: Done :) I like the check for `/etc/iis.version`.

micprog:
LGTM :+1:

PR #278, opened 2022-11-29
mem_to_banks: Move from axi to common_cells

thommythomaso:


micprog:
Can you update the dependency requirement in `axi.core` and `ips_list.yml` to common_cells 1.27.0 as well? Otherwise LGTM

PR #279, opened 2022-12-06
axi_pkg: Add functions which return channel widths

thommythomaso:


micprog (approved):
LGTM. AFAIK most of these things could be handled by $bits, but I think the reasoning here is that this is not supported properly by all tools.

PR #281, opened 2023-01-18
axi_dw_downsizer: initialize the write response buffer with EXOKAY

cfuguet:
This modification allows to correctly forward the response status of multi-beat exclusive write operations.

This RESP_EXOKAY response code has the lowest priority (as defined by the axi_pkg::resp_precedence function) when merging the response code from a multi-beat write response. Without this modification, a successful exclusive access can never get the corresponding RESP_EXOKAY code.

### DIFF
--- a/src/axi_dw_downsizer.sv
+++ b/src/axi_dw_downsizer.sv
@@ -397,11 +397,12 @@ module axi_dw_downsizer #(
               r_state_d = R_PASSTHROUGH;
 
               // Save beat
-              r_req_d.ar           = slv_req_i.ar     ;
-              r_req_d.ar_valid     = 1'b1             ;
-              r_req_d.burst_len    = slv_req_i.ar.len ;
-              r_req_d.orig_ar_size = slv_req_i.ar.size;
-              r_req_d.injected_aw  = 1'b0             ;
+              r_req_d.ar           = slv_req_i.ar        ;
+              r_req_d.ar_valid     = 1'b1                ;
+              r_req_d.burst_len    = slv_req_i.ar.len    ;
+              r_req_d.orig_ar_size = slv_req_i.ar.size   ;
+              r_req_d.injected_aw  = 1'b0                ;
+              r_req_d.r.resp       = axi_pkg::RESP_EXOKAY;
 
               case (r_req_d.ar.burst)
                 axi_pkg::BURST_INCR : begin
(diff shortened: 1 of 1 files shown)


cfuguet:
Yes, you are right. I can do the modifications for AR and R channels, too.

I will do that and push as soon as possible.

Thanks for the review

cfuguet:
Hello, I did the changes.

Regards,

src/axi_dw_downsizer.sv
```diff
@@ -397,11 +397,12 @@ module axi_dw_downsizer #(
...
-              r_req_d.ar_valid     = 1'b1             ;
-              r_req_d.burst_len    = slv_req_i.ar.len ;
-              r_req_d.orig_ar_size = slv_req_i.ar.size;
-              r_req_d.injected_aw  = 1'b0             ;
+              r_req_d.ar           = slv_req_i.ar        ;
+              r_req_d.ar_valid     = 1'b1                ;
+              r_req_d.burst_len    = slv_req_i.ar.len    ;
+              r_req_d.orig_ar_size = slv_req_i.ar.size   ;
+              r_req_d.injected_aw  = 1'b0                ;
+              r_req_d.burst_resp   = axi_pkg::RESP_EXOKAY;
```
SamuelRiedel, line 405: Thanks. :) However, this member does not exist in this struct. I think you have to set the default value here: https://github.com/pulp-platform/axi/blob/master/src/axi_dw_downsizer.sv#L383 with `r_req_d.r.resp = axi_pkg::RESP_EXOKAY;`

src/axi_dw_downsizer.sv
```diff
@@ -476,6 +477,7 @@ module axi_dw_downsizer #(
           r_req_d.burst_len    = w_req_q.orig_aw_len  ;
           r_req_d.orig_ar_size = w_req_q.orig_aw_size ;
           r_req_d.injected_aw  = 1'b1                 ;
+          r_req_d.burst_resp   = axi_pkg::RESP_EXOKAY ;
```
SamuelRiedel, line 480: Same here.

SamuelRiedel (changes requested):

SamuelRiedel (approved):
Thanks :+1: The internal CI also passes, so this looks good to me.

PR #290, opened 2023-03-28
axi_test: Towards file-based verification

thommythomaso:


micprog (commented):
Can you add some documentation on how to use this?

Issue #292, opened 2023-03-29
axi_pkg::LenWidth not compatible with vivado IP packager


CyrilKoe:
Since commit [fe4e020](https://github.com/pulp-platform/axi/commit/fe4e020f2c06a242d4901b23cedbf2eeab22bfe6) the typedef`axi_pkg::burst_t` now relies on a Verilog parameter `BurstWidth`.
https://github.com/pulp-platform/axi/blob/fd60be8b51a4fa7476856be162ce3334474592ba/src/axi_pkg.sv#L46
In [Occamy emulation](https://github.com/pulp-platform/snitch/blob/master/hw/system/occamy/src/occamy_xilinx.sv) we use these typedefs at the ports of a custom packaged Vivado IP.

But Vivado requires that "the IP ports are contained within the port description of the HDL file" (c.f. [doc](https://www.xilinx.com/support/documents/sw_manuals/xilinx2022_1/ug1118-vivado-creating-packaging-custom-ip.pdf) page 14).
So custom Vivado IPs can't use `axi_pkg::burst_t` at their top anymore (similar issue on this [support thread](https://support.xilinx.com/s/question/0D52E00006hpfh5SAA/ipflow-19627-hdl-parameter-fpgamanu-fpga-manu-xpath-expression-failed-undefined-parameter-xilinx-used-in-expression-xilinx?language=en_US))

We could maybe define the typdef with a numerical value and then the parameter based on the `$size` of this typedef.
See proposition in  https://github.com/pulp-platform/axi/commit/d2f98d691594c1a8dd1af5bedd5d37d965027535

PR #293, opened 2023-03-29
axi_pkg: keep hardcoded value in typedef for vivado ip packager

CyrilKoe:
\[ From issue https://github.com/pulp-platform/axi/issues/292 \]

Since commit [fe4e020](https://github.com/pulp-platform/axi/commit/fe4e020f2c06a242d4901b23cedbf2eeab22bfe6) the typedef`axi_pkg::burst_t` now relies on a Verilog parameter `BurstWidth`.
https://github.com/pulp-platform/axi/blob/fd60be8b51a4fa7476856be162ce3334474592ba/src/axi_pkg.sv#L46
In [Occamy emulation](https://github.com/pulp-platform/snitch/blob/master/hw/system/occamy/src/occamy_xilinx.sv) we use these typedefs at the ports of a custom packaged Vivado IP.

But Vivado requires that "the IP ports are contained within the port description of the HDL file" (c.f. [doc](https://www.xilinx.com/support/documents/sw_manuals/xilinx2022_1/ug1118-vivado-creating-packaging-custom-ip.pdf) page 14).
So custom Vivado IPs can't use `axi_pkg::burst_t` at their top anymore (similar issue on this [support thread](https://support.xilinx.com/s/question/0D52E00006hpfh5SAA/ipflow-19627-hdl-parameter-fpgamanu-fpga-manu-xpath-expression-failed-undefined-parameter-xilinx-used-in-expression-xilinx?language=en_US))

We could maybe define the typdef with a numerical value and then the parameter based on the `$size` of this typedef.
See proposition in  https://github.com/pulp-platform/axi/commit/d2f98d691594c1a8dd1af5bedd5d37d965027535

### DIFF
--- a/src/axi_pkg.sv
+++ b/src/axi_pkg.sv
@@ -19,49 +19,50 @@
 //! AXI Package
 /// Contains all necessary type definitions, constants, and generally useful functions.
 package axi_pkg;
-    /// AXI Transaction Burst Width.
-  parameter int unsigned BurstWidth  = 32'd2;
+
+  /// AXI Transaction Burst Width.
+  typedef logic [1:0]  burst_t;
+  /// AXI Transaction Response Type.
+  typedef logic [1:0]   resp_t;
+  /// AXI Transaction Cacheability Type.
+  typedef logic [3:0]  cache_t;
+  /// AXI Transaction Protection Type.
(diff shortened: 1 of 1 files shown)


CyrilKoe:
@thommythomaso should we re-define the value twice instead?

PR #297, opened 2023-04-09
Fix typo in comments

zeeshanrafique23:


micprog:
I don't think this is a typo, as the assertion aims to ensure the signals remain constant from the point where the valid bit is set to the point where the (ready-valid) handshake completes.

PR #312, opened 2023-07-20
Propagation of the cdc gray SYNC_STAGES parameter to the interfaces

mattsini1:
Here is a minor update to the CDCs.
The only changes are related to the propagation of the sync_stages parameter to the internal cdc_gray.
It is useful to set a different number of synchronization registers.


thommythomaso (approved):
Thanks for the contribution. Looks good to us. :+1:

PR #320, opened 2023-07-27
Properly handle response code in axi_to_detailed_mem

### LINKED ISSUE #319: ``axi_to_mem``: Error response signals in B and R response channels
I made the following observations regarding the assignment of the ``resp`` field of the B and R response structs in ``axi_to_detailed_mem``, lines https://github.com/pulp-platform/axi/blob/634b7438dddfc356fa1e5d9218ec314a78de09ea/src/axi_to_detailed_mem.sv#L502 and https://github.com/pulp-platform/axi/blob/634b7438dddfc356fa1e5d9218ec314a78de09ea/src/axi_to_detailed_mem.sv#L511

- ``mem_err_i`` and ``mem_exokay`` are possibly multi-bit, if ``NumBanks`` is not equal to 1. Therefore, I think either a vector-reduction ``OR`` or ``AND`` should be placed in front of those signals when deciding on the appropriate AXI response types.
- ``mem_err_i`` and ``mem_exokay``, and, in the same fashion possibly additional side-channel/meta signals from the downstream memory are used "directly" here, while the other fields of ``axi_resp_o.b/r`` are taken from buffers "deep" within the module. Particularly, ``m2s_resp`` in the ``R`` response comes out of a FIFO that can be any number of elements deep. So does this not create misalignment issues between the fields in the responses, e.g., an error response from a different read than what the data in ``m2s_resp`` belong to?
For a multi-beat AXI write burst where only one or few beats in the middle of the burst cause the memory to respond with an error, won't this cause the error to get lost and not being propagated to the B response at the end of the beat?
For multi-beat read bursts, if the response FIFO in ``stream_to_mem`` actually fills and cannot operate only in pass-through, this should similarly lead to error responses being assigned to the wrong R-response, or lost.

micprog:
Fixes #319

Requires https://github.com/pulp-platform/common_cells/pull/194 to be released

### DIFF
--- a/.github/workflows/gitlab-ci.yml
+++ b/.github/workflows/gitlab-ci.yml
@@ -4,10 +4,6 @@ on:
     branches-ignore:
       - gh-pages  # deployment target branch (this workflow should not exist on that branch anyway)
       - v**       # such branch names conflict with tags
-  pull_request:
-    branches-ignore:
-      - gh-pages  # deployment target branch (this workflow should not exist on that branch anyway)
-      - v**       # such branch names conflict with tags
   workflow_dispatch:
 
 jobs:
--- a/Bender.yml
+++ b/Bender.yml
@@ -19,7 +19,7 @@ package:
     - "Florian Zaruba <zarubaf@iis.ee.ethz.ch>"
 
 dependencies:
-  common_cells: { git: "https://github.com/pulp-platform/common_cells.git", version: 1.27.0 }
+  common_cells: { git: "https://github.com/pulp-platform/common_cells.git", rev: "be16fe6efa97af39ca258e398f37b7910e636e02" }
   common_verification: { git: "https://github.com/pulp-platform/common_verification.git", version: 0.2.3 }
   tech_cells_generic:  { git: "https://github.com/pulp-platform/tech_cells_generic.git", version: 0.2.2 }
 
--- a/Bender.yml
+++ b/Bender.yml
@@ -19,7 +19,7 @@ package:
     - "Florian Zaruba <zarubaf@iis.ee.ethz.ch>"
 
 dependencies:
-  common_cells: { git: "https://github.com/pulp-platform/common_cells.git", rev: "99fcffdc3f8a5f3f590f6c9f687f65a2e9f2e0fe" } # updated mem_to_banks_detailed
+  common_cells: { git: "https://github.com/pulp-platform/common_cells.git", version: 1.31.0 }
   common_verification: { git: "https://github.com/pulp-platform/common_verification.git", version: 0.2.3 }
   tech_cells_generic:  { git: "https://github.com/pulp-platform/tech_cells_generic.git", version: 0.2.2 }
 
--- a/Bender.yml
+++ b/Bender.yml
@@ -19,7 +19,7 @@ package:
     - "Florian Zaruba <zarubaf@iis.ee.ethz.ch>"
 
 dependencies:
-  common_cells: { git: "https://github.com/pulp-platform/common_cells.git", rev: "be16fe6efa97af39ca258e398f37b7910e636e02" }
+  common_cells: { git: "https://github.com/pulp-platform/common_cells.git", rev: "99fcffdc3f8a5f3f590f6c9f687f65a2e9f2e0fe" } # updated mem_to_banks_detailed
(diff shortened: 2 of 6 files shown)


.github/workflows/gitlab-ci.yml
thommythomaso, line 1: As a not. Lets not forget the changelog to be updated accordingly.

Bender.yml
```diff
@@ -19,7 +19,7 @@ package:
     - "Florian Zaruba <zarubaf@iis.ee.ethz.ch>"
 
 dependencies:
-  common_cells: { git: "https://github.com/pulp-platform/common_cells.git", version: 1.27.0 }
+  common_cells: { git: "https://github.com/pulp-platform/common_cells.git", rev: "be16fe6efa97af39ca258e398f37b7910e636e02" }
```
thommythomaso, line 22: Improved version of mem_to_banks.

src/axi_to_detailed_mem.sv
```diff
@@ -496,19 +518,50 @@ module axi_to_detailed_mem #(
...
 
+  // Accumulate errors if memory responds
+  localparam NumBytesPerBank = DataWidth/NumBanks/8;
+  logic collect_b_err_d, collect_b_err_q;
+  logic collect_b_exokay_d, collect_b_exokay_q;
+  logic [NumBanks-1:0] meta_buf_bank_strb, meta_buf_size_enable;
+  logic resp_b_err, resp_b_exokay, resp_r_err, resp_r_exokay;
+  for (genvar i = 0; i < NumBanks; i++) begin
+    assign meta_buf_bank_strb[i] = meta_buf.strb[i*NumBytesPerBank +: NumBytesPerBank];
+    assign meta_buf_size_enable[i] = i*NumBytesPerBank + NumBytesPerBank > (meta_buf.addr % DataWidth/8) &&
```
thommythomaso, line 529: Quite a full line. Can you maybe explain a bit in more details what is going on here?

thommythomaso (changes requested):
Can you comment a bit more what is going on exactly in the pretty dense code blocks added?

src/axi_to_detailed_mem.sv
glaserf: I agree, maybe at least one further level of parentheses would also help with understanding this rather complex condition.

src/axi_to_detailed_mem.sv
```diff
@@ -496,19 +518,50 @@ module axi_to_detailed_mem #(
...
   );
 
+  // Accumulate errors if memory responds
+  localparam NumBytesPerBank = DataWidth/NumBanks/8;
+  logic collect_b_err_d, collect_b_err_q;
+  logic collect_b_exokay_d, collect_b_exokay_q;
+  logic [NumBanks-1:0] meta_buf_bank_strb, meta_buf_size_enable;
+  logic resp_b_err, resp_b_exokay, resp_r_err, resp_r_exokay;
+  for (genvar i = 0; i < NumBanks; i++) begin
+    assign meta_buf_bank_strb[i] = meta_buf.strb[i*NumBytesPerBank +: NumBytesPerBank];
```
glaserf: ``meta_buf.strb`` is of type ``axi_strb_t`` but ``meta_buf_bank_strb`` is a 1-bit ``logic`` (for every bank).

glaserf (changes requested):

.github/workflows/gitlab-ci.yml
micprog: Added info in the changelog

Bender.yml
micprog: Updated `common_cells` to a proper version tag with the required changes

src/axi_to_detailed_mem.sv
micprog: I expanded the explanation to clarify what is going on and added an additional level of parentheses for clearer grouping.

src/axi_to_detailed_mem.sv
micprog: I also discovered this and appropriately fixed the issue, ensuring correct mapping:
```
    assign meta_buf_bank_strb[i] = |meta_buf.strb[i*NumBytesPerBank +: NumBytesPerBank];
```

thommythomaso (approved):

PR #322, opened 2023-09-17
Add reset states to axi_atop_filter to fix zero time sim hang

uge:
Runny occamy simulations in xsim sometimes triggers 0 time simulation loops. One loop seems to be related to use of atop_filter and the massaging of ready/valid signals to downstream axi library IP which may also have combinational paths depending on their spill parameterization. This change introduces a reset state to the both the read and write state machines which seems to eliminate one of the zero time loops in occamy simulation with xsim.

### DIFF
--- a/src/axi_atop_filter.sv
+++ b/src/axi_atop_filter.sv
@@ -67,11 +67,11 @@ module axi_atop_filter #(
   cnt_t   w_cnt_d, w_cnt_q;
 
   typedef enum logic [2:0] {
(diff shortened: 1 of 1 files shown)


micprog (approved):
LGTM, internal CI passed, does not look like it will affect the final HW too much, rebasing on master and merging.

micprog:
Closing as merged in https://github.com/pulp-platform/axi/commit/71ed21a906942be7608d294f88ad7f108a38a150

PR #329, opened 2024-01-16
Minor improvements to `axi_rand_master`

luca-valente:


src/axi_test.sv
```diff
@@ -808,11 +809,15 @@ package axi_test;
       mem_map.push_back({addr_begin, addr_end, mem_type});
     endfunction
 
-    function void add_traffic_shaping(input int unsigned len, input int unsigned freq);
+    function void clear_memory_regions();
+       mem_map.delete();
+    endfunction
+
+    function void add_traffic_shaping(input int unsigned len, input int unsigned size, input int unsigned freq);
```
micprog, line 816: This changes the interface making a non-breaking release difficult. @luca-valente do you think you can use a separate function (e.g. called `add_traffic_shaping_sized`) to add this functionality, and keeping the old function the same but wrapping this new one?

micprog (changes requested):
LGTM, if possible I would prefer to not change the interface though.

src/axi_test.sv
luca-valente: Sure! I can do that :smile: thanks for the feedback

luca-valente: @micprog I added a second function (`add_traffic_shaping_with_size`) and reverted the changes to `add_traffic_shaping` to avoid changing the interface. Let me know if that works for you. Thanks!

micprog (approved):
LGTM :+1:

PR #331, opened 2024-02-06
Fix byte calculation in axi_sim_mem R and W forks

LucaZulberti:
Hi all!

I found this bug when accessing the simulation RAM with `ax.size = 0` and 32-bit data bus.
During an increment access to the memory, when `r_cnt % strb_width != 0`, the address of the location accessed (`addr`) is calculated with `ax_queue[0].addr` (e.g. 2nd beat of an INCR burst at location 0x10 has an address of 0x11), but then `addr` is used to determine the byte in the 32-bit word, causing it to make the offset with 0x11 instead of 0x10.
In particular, for `r_cnt = 1`, the byte calculated was 2 instead of 1.

### DIFF
--- a/src/axi_sim_mem.sv
+++ b/src/axi_sim_mem.sv
@@ -162,8 +162,8 @@ module axi_sim_mem #(
             mon_w.user = aw_queue[0].user;
             mon_w.beat_count = w_cnt;
             for (shortint unsigned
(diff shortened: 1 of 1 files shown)


micprog (approved):
LGTM :+1:

PR #335, opened 2024-03-14
axi_from_reg: Add module

paulsc96:
Adds a simple Regbus to AXI4 adapter.

Valid Regbus requests are locked in until a response (B/R) is received. The data width must match on both ends. The address is truncated or zero-extended as required by the provided type parameters. AXI4 AX requests are sent with ID 0 and user signals 0. The cache signals are parameterizable for convenience.

The module was tested in Cheshire by using it to adapt the SpinalHDL USB control interface from AXI4 to Regbus. All tests pass and USB works as expected on FPGA.


micprog:
AFAIK register interface includes axi as a dependency. Therefore I would move this module there, as this one requires both reg and axi types.

paulsc96:
can do.

PR #338, opened 2024-04-30
`axi_sim_mem`: Configure response data

micprog:
Allow response data for uninitialized region to have configurable defined value, based on a parameter.


thommythomaso (approved):
LGTM :+1:

PR #343, opened 2024-07-03
Add `axi_xbar_unmuxed`: Partial crossbar with unmultiplexed mst_ports

micprog:
Adds new module with contents from `axi_xbar`, replaces contents of `axi_xbar` with module instantiation to ensure single code source.


thommythomaso (approved):
:+1:

PR #350, opened 2024-07-29
axi_xbar_unmuxed: Disable for VCS compilation

fischeti:
multi-dimensional array interfaces are not yet supported in VCS


micprog (approved):
LGTM :+1:

PR #358, opened 2024-11-13
VCS and Verilator support

LucaRufer:
Minimal changes required to compile sources and simulation targets for VCS (version W-2024.09-1) and Verilator (v5.028-44-g1d79f5c59):
- Verilator requires the `std::` prefix when using built-in classes like `std::semaphore` and `std::mailbox`.
- Verilator cannot compute static variables at compile time when conditional operator is used.
- Verilator cannot determine the default type for Associative Arrays if a parameter is used.
- VCS requires parameter types to match when calling a function with a class object. If any of the axi_??_beat classes are instantiated directly and not from the typedefs in axi_driver, they cannot be passed to a function of a axi_driver object, as `parameter` and `parameter int` are not the same.

### DIFF
--- a/src/axi_sim_mem.sv
+++ b/src/axi_sim_mem.sv
@@ -116,8 +116,8 @@ module axi_sim_mem #(
 
   monitor_t [NumPorts-1:0] mon_w, mon_r;
   logic [7:0]     mem[addr_t];
-  axi_pkg::resp_t rerr[addr_t] = '{default: axi_pkg::RESP_OKAY};
-  axi_pkg::resp_t werr[addr_t] = '{default: axi_pkg::RESP_OKAY};
+  axi_pkg::resp_t rerr[addr_t] = '{default: 2'b0};
+  axi_pkg::resp_t werr[addr_t] = '{default: 2'b0};
 
   // error happened in write burst
   axi_pkg::resp_t [NumPorts-1:0] error_happened = axi_pkg::RESP_OKAY;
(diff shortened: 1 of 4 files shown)


src/axi_sim_mem.sv
thommythomaso, line 1: Which tool is upset with using the values from the package?

thommythomaso (commented):
Thanks for the fixes. Can you change the parameters to `int unsigned` directly?
Also, why don't we reference the package anymore for the AXI response in the simulation memory?

src/axi_sim_mem.sv
LucaRufer: Verilator cannot compile with the following error:
```
%Error: Internal Error: src/axi_sim_mem.sv:119:34: ../V3Width.cpp:7074: Node has no type:
  119 |   axi_pkg::resp_t rerr[addr_t] = '{default: axi_pkg::RESP_OKAY};
      |                                  ^~
                        ... See the manual at https://verilator.org/verilator_doc.html for more assistance.
```
I tried different ways of re-writing using casting and other approaches, but it just does not work with Verilator.

thommythomaso: Ok, can you then keep the link to the package as a comment and write a note why this was changed? Please open an issue for that on the verilator page to have this eventually fixed in the simulator.

LucaRufer:
> Thanks for the fixes. Can you change the parameters to `int unsigned` directly? Also, why don't we reference the package anymore for the AXI response in the simulation memory?

I updated the pull request as requested. 
Verilator has problems with typed defaults for arrays. Even casting does not work. The only way I found to make it compilable with Verilator is to expand it to zero.

thommythomaso (approved):

PR #369, opened 2025-03-18
Bug fixing for axi_burst_unwrap module

Aquaticfuller:
1. Fix wrap len and boundary logic

This commit fixes an off‐by‐one error in the burst length calculation for wrapping bursts in the AXI interface. According to the AXI specification, the actual burst length is encoded as (AxLEN + 1). The fix adjusts the container size calculation by changing it from `ax_i.len << ax_i.size` to `(ax_i.len + 1) << ax_i.size` and modifies the computed burst lengths accordingly.

2. When an AR wrap burst comes in but master ready is low, don't increase the axi_burst_counter, or the unhandshaked AR will be dropped.

### DIFF
--- a/src/axi_burst_unwrap.sv
+++ b/src/axi_burst_unwrap.sv
@@ -474,7 +474,7 @@ module axi_burst_unwrap_ax_chan #(
 
   enum logic {Idle, Busy} state_d, state_q;
   always_comb begin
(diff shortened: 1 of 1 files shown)


niwis (approved):
Thanks, LGTM

thommythomaso (approved):

PR #371, opened 2025-03-26
Axi_cut: Add selective channel bypassing

Burinir:
Add the parameters  BypassAw, BypassW, BypassB, BypassAr, BypassR which are by default all assigned to the already existing parameter Bypass. This allowes selectively bypassing the cut on certain channels. This is useful to only cut the channels that lead to timing issues.


micprog (approved):
LGTM :+1:

PR #372, opened 2025-04-11
axi_xbar_unmuxed: judge the connection relationship before routing

Alvin-FA:
For demux, we can check the connectivity before routing, this can improve transmission efficiency. The more there are such case that connectivity[i][j] == 0 between slave_port[i] and mst_port[j], the better the improvement effect will be. Because the unnecessary arbitrations have been reduced, this accelerates the correct access response. At the same time, ensure that there is a dedicated port to receive decode errors as before.

### DIFF
--- a/src/axi_xbar_unmuxed.sv
+++ b/src/axi_xbar_unmuxed.sv
@@ -104,16 +104,31 @@ import cf_math_pkg::idx_width;
 
   // workaround for issue #133 (problem with vsim 10.6c)
   localparam int unsigned cfg_NoMstPorts = Cfg.NoMstPorts;
+  // addr_map_i and connectivity relationship
+  rule_t [Cfg.NoSlvPorts-1:0][Cfg.NoAddrRules-1:0]  addr_map;
 
   for (genvar i = 0; i < Cfg.NoSlvPorts; i++) begin : gen_slv_port_demux
-`ifdef VCS
-    logic [MstPortsIdxWidth-1:0]          dec_aw,        dec_ar;
-`else
-    logic [idx_width(Cfg.NoMstPorts)-1:0] dec_aw,        dec_ar;
-`endif
+    `ifdef VCS
+        logic [MstPortsIdxWidth-1:0]          dec_aw,        dec_ar;
+    `else
+        logic [idx_width(Cfg.NoMstPorts)-1:0] dec_aw,        dec_ar;
+    `endif
+
     mst_port_idx_t                        slv_aw_select, slv_ar_select;
     logic                                 dec_aw_valid,  dec_aw_error;
     logic                                 dec_ar_valid,  dec_ar_error;
+
+    //if there is no connection betwen master and slave, one index for decode error will be used
+    for (genvar j = 0; j < Cfg.NoMstPorts; j++) begin : gen_addr_map
+      if (!Connectivity[i][j]) begin : fix_addr_map
+        assign addr_map[i][j].idx         = mst_port_idx_t'(Cfg.NoMstPorts);
+        assign addr_map[i][j].start_addr  = addr_map_i[j].start_addr;
+        assign addr_map[i][j].end_addr    = addr_map_i[j].end_addr;
+      end
+      else begin : keep_addr_map
+        assign addr_map[i][j] = addr_map_i[j];
+      end
+    end
 
     addr_decode #(
       .NoIndices  ( Cfg.NoMstPorts  ),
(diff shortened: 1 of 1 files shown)


micprog (commented):
This looks like an interesting improvement! I have one concern: What happens if a specific port has `en_default_mst_port_i` set and the `default_mst_port_i` for that slv_port is not connected? Let's double-check that we can handle this scenario correctly as well.

Alvin-FA:
> This looks like an interesting improvement! I have one concern: What happens if a specific port has `en_default_mst_port_i` set and the `default_mst_port_i` for that slv_port is not connected? Let's double-check that we can handle this scenario correctly as well.
I think this situation belongs to the error in writing the configuration file

Alvin-FA:
> This looks like an interesting improvement! I have one concern: What happens if a specific port has `en_default_mst_port_i` set and the `default_mst_port_i` for that slv_port is not connected? Let's double-check that we can handle this scenario correctly as well.

I think this scenario belongs to a configuration file error. If someone use en_default_mst_port_i and default_mst_port_i, he/she should ensure the connectivity correctly. Unfortunately, if this scenario really occurs, we can only ensure that the route to the instantiation of axi_err_slv module to avoid deadlock.

micprog:
> > This looks like an interesting improvement! I have one concern: What happens if a specific port has `en_default_mst_port_i` set and the `default_mst_port_i` for that slv_port is not connected? Let's double-check that we can handle this scenario correctly as well.
> 
> I think this scenario belongs to a configuration file error. If someone use en_default_mst_port_i and default_mst_port_i, he/she should ensure the connectivity correctly. Unfortunately, if this scenario really occurs, we can only ensure that the route to the instantiation of axi_err_slv module to avoid deadlock.

I agree, if this occurs, then whoever set this up made a mistake. Nonetheless, we can still support this case, maybe with a parameter as an option to disable it?

I think we could directly handle this by setting the corresponding `default_idx_i` of the `addr_decode` modules as follows:

```diff
-      .default_idx_i ( default_mst_port_i[i] )
+      .default_idx_i ( Connectivity[i][default_mst_port_i[i]] ? default_mst_port_i[i] : mst_port_idx_t'(Cfg.NoMstPorts) )
```

There may be some PPA implications in non-static cases, but this approach should still properly return DECERR for routing to an unconnected region.

Alvin-FA:
> > > This looks like an interesting improvement! I have one concern: What happens if a specific port has `en_default_mst_port_i` set and the `default_mst_port_i` for that slv_port is not connected? Let's double-check that we can handle this scenario correctly as well.
> > 
> > 
> > I think this scenario belongs to a configuration file error. If someone use en_default_mst_port_i and default_mst_port_i, he/she should ensure the connectivity correctly. Unfortunately, if this scenario really occurs, we can only ensure that the route to the instantiation of axi_err_slv module to avoid deadlock.
> 
> I agree, if this occurs, then whoever set this up made a mistake. Nonetheless, we can still support this case, maybe with a parameter as an option to disable it?
> 
> I think we could directly handle this by setting the corresponding `default_idx_i` of the `addr_decode` modules as follows:
> 
> ```diff
> -      .default_idx_i ( default_mst_port_i[i] )
> +      .default_idx_i ( Connectivity[i][default_mst_port_i[i]] ? default_mst_port_i[i] : mst_port_idx_t'(Cfg.NoMstPorts) )
> ```
> 
> There may be some PPA implications in non-static cases, but this approach should still properly return DECERR for routing to an unconnected region.

Excellent! It is a very ingenious solution.

src/axi_xbar_unmuxed.sv
```diff
@@ -104,25 +104,40 @@ import cf_math_pkg::idx_width;
...
+    for (genvar j = 0; j < Cfg.NoMstPorts; j++) begin : gen_addr_map
+      if (!Connectivity[i][j]) begin : fix_addr_map
+        assign addr_map[i][j].idx         = mst_port_idx_t'(Cfg.NoMstPorts);
+        assign addr_map[i][j].start_addr  = addr_map_i[j].start_addr;
+        assign addr_map[i][j].end_addr    = addr_map_i[j].end_addr;
+      end
+      else begin : keep_addr_map
+        assign addr_map[i][j] = addr_map_i[j];
+      end
+    end
```
micprog, line 131: I think there is a major issue here when your `Cfg.NoMstPorts` does not equal your `Cfg.NoAddrRules`, and if they are not sorted the same. I suggest you adapt this section to iterate over all address map entries and check for connectivity of the targeted id, e.g., as below:
```suggestion
    for (genvar j = 0; j < Cfg.NoAddrRules; j++) begin : gen_addr_map
      always_comb begin
        addr_map[i][j] = addr_map_i[j];
        if (!Connectivity[i][addr_map_i[j].idx]) begin
          addr_map[i][j].idx = mst_port_idx_t'(Cfg.NoMstPorts);
        end
      end
    end
```

Alvin-FA: Thank you for this suggestion! This made me realize my deficiencies.

PR #373, opened 2025-05-20
axi_sim_mem: propagate `user` field on B and R channels

ricted98:


micprog (approved):
I don't think that there is a clear guidance or expected behavior the user signals should have and there can be discussions about what behavior makes the most sense. Nonetheless, I think this behavior is as good as any, so I'm fine accepting and integrating this.

PR #382, opened 2025-07-01
Only invalidate WRAP bursts if they are unmodifiable

Scheremo:
This PR changes the `axi_burst_unwrap` module to only return a slave error if a WRAP burst transaction is marked unmodifiable; before, all unmodifiable transactions returned a slave error, even the ones that the module would not have modified.

### DIFF
--- a/src/axi_burst_unwrap.sv
+++ b/src/axi_burst_unwrap.sv
@@ -103,7 +103,6 @@ module axi_burst_unwrap #(
     // The AXI Spec (A3.4.1) only allows splitting non-modifiable transactions ..
     // ... but this module only splits WRAP bursts, so ignore all others
     if (!axi_pkg::modifiable(cache) && (burst == axi_pkg::BURST_WRAP)) begin
-      // .. if they are INCR bursts and longer than 16 beats.
       return 1'b0;
     end
     // All other transactions are supported.


src/axi_burst_unwrap.sv
```diff
@@ -101,9 +101,10 @@ module axi_burst_unwrap #(
     // ATOPs are not supported.
     if (atop != '0) return 1'b0;
     // The AXI Spec (A3.4.1) only allows splitting non-modifiable transactions ..
-    if (!axi_pkg::modifiable(cache)) begin
+    // ... but this module only splits WRAP bursts, so ignore all others
+    if (!axi_pkg::modifiable(cache) && (burst == axi_pkg::BURST_WRAP)) begin
       // .. if they are INCR bursts and longer than 16 beats.
```
micprog, line 106:
```suggestion
```
I think with this change the comment here is now very misleading.

Scheremo: Hey Michael! Leave it to me to make a mistake in a 5 line PR :)
I removed the comment in the [latest commit](https://github.com/pulp-platform/axi/pull/382/commits/a459f058ed5540d8b660aca909fb1dfeb808dd2d).

niwis (approved):
LGTM. I think we *may* allow splitting non-modifiable wrapping bursts with a length greater than 16; at least I can't find a restriction to `INCR` bursts in issue K (A5.2.2)

Scheremo:
> LGTM. I think we _may_ allow splitting non-modifiable wrapping bursts with a length greater than 16; at least I can't find a restriction to `INCR` bursts in issue K (A5.2.2)

Hi Nils, if I understand the spec correctly the maximum length for WRAP bursts is 16 (A4.1.4).

niwis:
Good point! Never mind, then :)

PR #387, opened 2025-07-23
axi_burst_unwrap: Remove overly pessimistic assertion

Scheremo:
This PR fixes a bug in the `axi_burst_unwrap` module, where the `id_queue` managing a write burst's outstanding W beats receives decrement requests even if the W channel transfer was not handshaked.

### DIFF
--- a/src/axi_burst_unwrap.sv
+++ b/src/axi_burst_unwrap.sv
@@ -176,6 +176,7 @@ module axi_burst_unwrap #(
     unique case (w_state_q)
       WReady: begin
         if (act_req.w_valid) begin
+          w_cnt_req = 1'b1;
           if (w_cnt_gnt) begin
             w_last_d = act_req.w.last | (w_cnt_len == 8'd0);
             mst_req_o.w.last  = w_last_d;
@@ -183,7 +184,6 @@ module axi_burst_unwrap #(
             // Try to forward the beat downstream.
             mst_req_o.w_valid = 1'b1;
             if (mst_resp_i.w_ready) begin
-              w_cnt_req = 1'b1;
               act_resp.w_ready = 1'b1;
               if (w_last_d && !act_req.w.last) begin
                 w_state_d = WFeedthrough;
--- a/src/axi_burst_unwrap.sv
+++ b/src/axi_burst_unwrap.sv
@@ -647,11 +647,4 @@ module axi_burst_counters #(
   // registers
   `FFARN(err_q, err_d, '0, clk_i, rst_ni)
 
-  `ifndef VERILATOR
-  // pragma translate_off
-  assume property (@(posedge clk_i) idq_oup_gnt |-> idq_oup_valid)
-    else $warning("Invalid output at ID queue, read not granted!");
-  // pragma translate_on
-  `endif
-
 endmodule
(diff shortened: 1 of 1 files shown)


Scheremo:
I think the same technical bug is in the handling of B channel responses, but it practically never triggers because most AXI modules in this repo always keep `b_ready` high. If we can get this change reviewed for the W channel, I can also fix it for the B channel.

Scheremo:
Hey Nils, You're right about the loop between `req` and `gnt`, I didn't notice this before.
 I'll check your proposed solution on my end, but it should be "more correct" :).

Scheremo:
Okay, so this looked better to begin with, until I figured out the `real real` issue at hand.
Since `w_cnt_req` is only gated by `w_valid`, it might be issued when a master is trying (unsuccessfully) to launch a new W transaction, i.e. `aw_valid` and `w_valid` are both high, but `aw_ready` is not. In this case, the assertion further down in the code triggers, since the id_queue gnt is given, without the data being valid, because the `w_valid` used for issuing the request stems from a transaction that has not yet been handshaked. 
I'm thinking maybe instead of 
```
      WReady: begin
        if (act_req.w_valid) begin
```
We need 
```
      WReady: begin
        if (act_req.w_valid & (w_cnt_len > 0 )) begin
```

Since the id_queue takes a cycle to latch a new transaction anyway. What do you think, @niwis ?

Scheremo:
> Okay, so this looked better to begin with, until I figured out the `real real` issue at hand. Since `w_cnt_req` is only gated by `w_valid`, it might be issued when a master is trying (unsuccessfully) to launch a new W transaction, i.e. `aw_valid` and `w_valid` are both high, but `aw_ready` is not. In this case, the assertion further down in the code triggers, since the id_queue gnt is given, without the data being valid, because the `w_valid` used for issuing the request stems from a transaction that has not yet been handshaked. I'm thinking maybe instead of
> 
> ```
>       WReady: begin
>         if (act_req.w_valid) begin
> ```
> 
> We need
> 
> ```
>       WReady: begin
>         if (act_req.w_valid & ("We are not in a half-way AW handshake" )) begin
> ```
> 
> Since the id_queue takes a cycle to latch a new transaction anyway. What do you think, @niwis ?
> 
> EDIT: Alternatively I think we can also disable the assertion further down in the code, as this case seems valid to me.

I think I solved the underlying issue now; by introducing an `cnt_empty_o` signal to the burst counters, this PR avoids requesting burst counts of outstanding write transactions iff no write transactions are in flight. 

If I understand correctly, functionally this only causes a change if the AXI slave of this module handshakes `w_ready` **without a write transaction in flight**, as in this case invalid data might be passed through. This is illegal behaviour as far as I interpret the AXI spec, so this should not change functionality when connected to a valid AXI slave.

The edge case this change prevents is the w counter being requested if there are no more outstanding transactions queued up, and the AW Channel handshake for the next transaction has not been done, but the `w_valid` for the first beat of the next transaction on the W channel is already asserted. Adding this change helps avoid the warning `"Invalid output at ID queue, read not granted!"`, since now the FSM awaits that the w counter is non-empty.

Scheremo:
I agree with @niwis on this - all preconditions for forwarding W beats are correct in the current version. The assertion, however, is overly pessimistic.
As the minimal change to get to a state where no erroneous assertions are triggered, we should remove the assertion rather than change the module's behaviour.

niwis (approved):

PR #388, opened 2025-07-24
axi_demux_simple: expose B and R arbiters indexes

ricted98:
For ACE related modifications to propagate WACK/RACK signals, it is useful to know which R or B beat has won the internal arbitration. I would propose to make the indexes externally visible. While the change is minimal, it has the drawback to change the interface of this module. What do you think?

### DIFF
--- a/src/axi_demux_simple.sv
+++ b/src/axi_demux_simple.sv
@@ -63,7 +63,9 @@ module axi_demux_simple #(
   output axi_resp_t                     slv_resp_o,
   // Master Ports
   output axi_req_t    [NoMstPorts-1:0]  mst_reqs_o,
-  input  axi_resp_t   [NoMstPorts-1:0]  mst_resps_i
+  input  axi_resp_t   [NoMstPorts-1:0]  mst_resps_i,
+  output select_t                       mst_b_idx_o,
+  output select_t                       mst_r_idx_o
 );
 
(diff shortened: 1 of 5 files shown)


micprog (approved):
This is a breaking change as calling modules will require an update, but I do not see any major issues.

src/axi_demux_simple.sv
```diff
@@ -63,7 +63,9 @@ module axi_demux_simple #(
   output axi_resp_t                     slv_resp_o,
   // Master Ports
   output axi_req_t    [NoMstPorts-1:0]  mst_reqs_o,
-  input  axi_resp_t   [NoMstPorts-1:0]  mst_resps_i
+  input  axi_resp_t   [NoMstPorts-1:0]  mst_resps_i,
+  output select_t                       mst_b_idx_o,
```
imchenwu, line 67: I would suggest adding the following instruction.
```suggestion
  // Only valid when the corresponding resp is valid
  output select_t                       mst_b_idx_o,
```
It would also be great to explain the intention of exposing these two ports here.

imchenwu (changes requested):
Exposing these two ports is fine to me as well. However, I would suggest some further changes:

1. When `NoMstPorts == 1`, it seems like there's no driver for `mst_b_idx_o`/`mst_r_idx_o`.
2. `select_t` is not strictly the same width as `b_idx/r_idx` (when `NoMstPorts==1`)

PR #390, opened 2025-07-25
Add an isolatable CDC module

LucaRufer:
Motivation: The current implementation of the `axi_cdc` module has the limitation that it does not allow for one-sided 'warm' resets of the module. This means that the `axi_cdc` module can "only" be used as a clock domain crossing, but not as a reset domain or power domain crossing with one-sided resets/power-downs.

To get rid of this limitation, this PR adds an isolatable CDC module `axi_cdc_isolatable`. This new module allows for one-sided resets or power-downs using the isolation signals available on both sides of the CDC crossing. These signals can be used to gracefully separate the two sides of the CDC and reset or turn them off individually, without facing the problem of the CDC FIFO pointers only being reset on one sideb only. Details on these signals can be found in the module description.

The PR adds the following modules:
`axi_cdc_isolatable`: Isolatable variant of the `axi_cdc`, allowing for one-sided resets or power-downs.
`axi_cdc_src_clearable`: Clearable variant of the 'axi_cdc_src` using the clearable CDC FIFO instead of the 'normal' CDC FIFO
`axi_cdc_dst_clearable`: Clearable variant of the 'axi_cdc_dst` using the clearable CDC FIFO instead of the 'normal' CDC FIFO

Additionally, the PR adds a testbench for the `axi_cdc_isolatable` module, which was adapted from the `axi_cdc` testbench. The testbench tests the one-sided resets for the src and the dst side.

### DIFF
--- a/.ci/Memora.yml
+++ b/.ci/Memora.yml
@@ -62,6 +62,21 @@ artifacts:
     outputs:
       - build/axi_cdc-%.tested
 
+  axi_cdc_isolatable-%:
+    inputs:
+      - Bender.yml
+      - include
+      - scripts/run_vsim.sh
+      - src/axi_pkg.sv
+      - src/axi_intf.sv
+      - src/axi_test.sv
+      - src/axi_cdc_dst_clearable.sv
+      - src/axi_cdc_src_clearable.sv
+      - src/axi_cdc_isolatable.sv
+      - test/tb_axi_cdc_isolatable.sv
+    outputs:
+      - build/axi_cdc_isolatable-%.tested
+
   axi_delayer-%:
     inputs:
       - Bender.yml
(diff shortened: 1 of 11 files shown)


micprog:
I'm a bit confused as to why we need to use a clearable cdc fifo for this implementation and cannot use the existing cdc fifo. In effect, the reset boundary does not need to equal the clock boundary, so you should be able to use the isolate module to properly ensure the slave is not active and set the reset boundary at either side of the CDC, ensuring both sides of the cdc fifo are reset together (or both sides of the cdc fifo are not reset when the slave is reset). This is similar to what we implemented for [Carfield](https://github.com/pulp-platform/carfield/blob/main/hw/cheshire_wrap.sv#L362-L419). Please let me know if I'm missing something!

Issue #226, opened 2022-04-04
Support Xilinx Vivado Simulator


skokvermon:
Hi

I tried to simulate few axi benchs using xilinx vivado simulator.
I could parse all the code with xvlog succesfuly
But during elaboration I get errors like:
```
$ xelab tb_axi_dw_downsizer
Vivado Simulator v2021.2
Copyright 1986-1999, 2001-2021 Xilinx, Inc. All Rights Reserved.
Running: /home/skok/tools/Xilinx/Vivado/2021.2/bin/unwrapped/lnx64.o/xelab tb_axi_dw_downsizer 
Multi-threading is on. Using 6 slave threads.
Starting static elaboration
Pass Through NonSizing Optimizer
ERROR: [VRFC 10-2991] 'randomize' is not declared under prefix 'w_beat' [/home/skok/src/axi/src/axi_test.sv:1196]
ERROR: [VRFC 10-2991] 'randomize' is not declared under prefix 'r_beat' [/home/skok/src/axi/src/axi_test.sv:1335]
ERROR: [VRFC 10-2991] 'randomize' is not declared under prefix 'b_beat' [/home/skok/src/axi/src/axi_test.sv:1387]
WARNING: [VRFC 10-3091] actual bit length 1 differs from formal bit length 2 for port 'req_i' [/home/skok/src/axi/src/axi_dw_downsizer.sv:146]
WARNING: [VRFC 10-3091] actual bit length 1 differs from formal bit length 2 for port 'gnt_o' [/home/skok/src/axi/src/axi_dw_downsizer.sv:147]
ERROR: [XSIM 43-3322] Static elaboration of top level Verilog design unit(s) in library work failed.
```

for another bench, there is a feature missing:
```
$ xelab tb_axi_delayer
Vivado Simulator v2021.2
Copyright 1986-1999, 2001-2021 Xilinx, Inc. All Rights Reserved.
Running: /home/skok/tools/Xilinx/Vivado/2021.2/bin/unwrapped/lnx64.o/xelab tb_axi_delayer 
Multi-threading is on. Using 6 slave threads.
Starting static elaboration
Pass Through NonSizing Optimizer
Completed static elaboration
ERROR: [XSIM 43-3980] File "/home/skok/src/axi/src/axi_test.sv" Line 865 : The SystemVerilog feature "implication operator" is not supported yet for simulation.
```
xilinx vivado simulator may not implement all systemverilog features, but it is free. And it works under linux and windows.
[...]

PR #391, opened 2025-07-29
Use XSIM define to remove all default disable blocks under src directory

jmdeharor:
Some files already have the XSIM ifndef to remove the `default disable` blocks (https://github.com/pulp-platform/axi/blob/master/src/axi_xbar_unmuxed.sv), which are not supported by XSIM. This MR completes this task to make all files consistent, and ensuring they can be simulated with Vivado.

### DIFF
--- a/src/axi_burst_splitter_gran.sv
+++ b/src/axi_burst_splitter_gran.sv
@@ -397,6 +397,7 @@ module axi_burst_splitter_gran #(
   // Assumptions and assertions
   // --------------------------------------------------
   `ifndef VERILATOR
+  `ifndef XSIM
   // pragma translate_off
   default disable iff (!rst_ni);
   // Inputs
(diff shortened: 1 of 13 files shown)


cousteaulecommandant:
I got this issue too.  It results in the error message `The SystemVerilog feature "Default Disable iff declaration" is not supported yet for simulation.` when trying to simulate in Vivado.  Apparently, XSIM doesn't support `default disable iff` constructs.  (As the original poster mentioned, these constructs are already disabled in *some* files for XSIM as well as Verilator, but only for Verilator on other files.  This PR would fix that.)

For now, the workaround I had to use was to add `VERILATOR` to the list of macro definitions in the simulation settings of Vivado in my local project.

PS: [This comment](https://github.com/pulp-platform/axi/issues/226#issuecomment-2952720481) suggests that newer versions of Vivado need to use the macro `XILINX_SIMULATOR` rather than `XSIM`.  Maybe all three should be checked?

PR #398, opened 2025-09-22
Add multicast support to XBAR

colluca:
This PR adds multicast support to the AXI XBAR, such that write requests from a master can be multicast to multiple slaves.

To support multicast with low area overhead, we adopt the mask-based encoding proposed in https://arxiv.org/pdf/2502.19215.

In detail, this PR implements the following changes:
- Adds the `any_outstanding_trx_o` port to the `axi_demux_id_counters` module.
- The `axi_mcast_demux_simple` implements a superset of the functionality of the `axi_demux_simple`, including all multicast capabilities, which require a few more signals on the interface, namely `mst_is_mcast_o` and `mst_aw_commit_o`. So `axi_demux_simple` is rewritten as a wrapper around `axi_mcast_demux_simple`, tying off unused ports and parameters.
- The same is done for the `axi_mcast_mux` and `axi_mux` modules, the `axi_mcast_xbar_unmuxed` and `axi_xbar_unmuxed` modules, and the `axi_mcast_xbar` and `axi_xbar` modules.
- The `axi_demux_id_counters` module is moved to its own file, which is anyways good practice, and it would otherwise not be clear whether it should now belong to `axi_demux_simple.sv` or `axi_mcast_demux_simple.sv`.
- Placing the spill registers downstream of the address decoders requires registering the `ax_select` signals from the address decoders. To eliminate this overhead (in the case of multicast it would even increase due to additional output signals from the decoders), we move the address decoders downstream of the spill registers in the`axi_mcast_demux_mapped` module. The `_mapped` prefix is to identify a demux whose slaves are memory-mapped, i.e. that uses address decoders to generate the `ax_select` signals. Since several modules use the demux with a non-address-based `ax_select` signal generation, the `axi_demux` module preserves the same interface and functionality as before, and does not wrap the `axi_mcast_demux_mapped` module, unlike the other unicast and multicast IP pairs.

Auxiliary changes:
- The `axi_demux_id_counters` is moved to its own file, as good practice.
- Update IP levels in `Bender.yml` after IP refactoring (including fixes to some dependencies that were not previously tracked, e.g. `axi_demux_simple`->`axi_burst_splitter_gran`, and others)

Additional notes:
- The definition of _default port_ should be adapted for multicast requests: "a port the request should be (also) forwarded to if any of the (multiple) addresses carried by the multicast request falls outside of the union of all the ports' address maps". For unicast requests it's cheap to determine if the address falls within the union of the address rules, by simply iterating over the address rules. For multicast requests this would be more complicated/expensive, so we opted for separately providing the union through the default master port rule (as an example, see https://github.com/pulp-platform/snitch_cluster/blob/6248ae825d821dca81c7b9c0f3cf3fd59c0ecdaa/hw/snitch_cluster/src/snitch_cluster.sv#L775-L782).
Details on how the default master port rule is used are in the `multiaddr_decode` module (see also https://github.com/pulp-platform/common_cells/pull/235).


src/axi_mcast_mux.sv
```diff
@@ -0,0 +1,665 @@
...
+    assign mcast_aw_priority = mcast_aw_valid && !lock_unicast_q;
+    assign mcast_aw_ready = aw_ready && !mcast_aw_commit &&  mcast_aw_priority;
+    assign ucast_aw_ready = aw_ready && !mcast_aw_commit && !mcast_aw_priority;
+    assign mst_aw_chan = mcast_aw_commit ? mcast_aw_chan : ucast_aw_chan;
+    assign slv_aw_readies = mcast_aw_readies | ucast_aw_readies;
+    // In case of a multicast transaction, downstream valid should only be asserted
+    // on commit. However, a multicast transaction is already accepted on the cycle
+    // before, when we have mcast_aw_valid. We must prevent a unicast transaction
+    // from raising the downstream valid in that time (with !mcast_aw_valid).
+    assign aw_valid = mcast_aw_commit || (ucast_aw_valid && !mcast_aw_valid);
```
micprog: This is the main additional logic added for multicast. Tracing back the logic it seems to keep the existing functionality, given the mcast inputs are tied to '0.

micprog (changes requested):
This PR is quite cumbersome to review, as most of the 5k line changes are code movement changes, and the code changes are not directly visible in the interface here. For some files, I added the corresponding git commands I used to check the changes. It might make sense to move some of these changes not related to multicast to dedicated PRs, simplifying the review and traceability of changes.

A further complication in reviewing this PR is the non-standard feature, as multicast is **not supported in any available AXI specification**. Please add documentation in the corresponding README files for what multicast is supported, how it works, and how it is intended to be used.

Also, given that this is not part of the spec, it might make sense to put this in a separate project, keeping this axi repository to what is available in the spec, but we can have that discussion offline.

Primarily, I was under the impression that this PR **does not change** the existing functionality when multicast is disabled or the existing modules are used, but this does not seem to be the case! 

Along with comments placed to help understanding, there are a few other items that need to be addressed. Please see below.

Bender.yml
```diff
@@ -19,7 +20,7 @@ package:
     - "Florian Zaruba <zarubaf@iis.ee.ethz.ch>"
 
 dependencies:
-  common_cells: { git: "https://github.com/pulp-platform/common_cells.git", version: 1.37.0 }
+  common_cells:        { git: "https://github.com/pulp-platform/common_cells.git", rev: "multicast-xbar" }
```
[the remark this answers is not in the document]
colluca: Unfortunately this is waiting on https://github.com/pulp-platform/common_cells/pull/235 (>1y...).

It's been marked as v2, but this should be hopefully sorted out shortly.

src/axi_pkg.sv
```diff
@@ -519,6 +519,12 @@ package axi_pkg;
     /// Each master port can have multiple rules, should have however at least one.
     /// If a transaction can not be routed the xbar will answer with an `axi_pkg::RESP_DECERR`.
     int unsigned   NoAddrRules;
+    /// The number of address rules to be considered for multicasting,
+    /// assumed to be at the start of `addr_map_i`.
+    int unsigned   NoMulticastRules;
+    /// Number of master ports of the crossbar which can be targets of a multicast request.
+    /// These are assumed to be connected at the lower indices.
+    int unsigned   NoMulticastPorts;
```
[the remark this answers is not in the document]
colluca: I think we agree on this point, but then how should we address it?

src/axi_mcast_xbar.sv
```diff
@@ -0,0 +1,265 @@
...
+  typedef logic [Cfg.AxiDataWidth       -1:0] data_t;
+  typedef logic [Cfg.AxiDataWidth/8     -1:0] strb_t;
+  typedef logic [AXI_USER_WIDTH         -1:0] user_t;
+  // AW channel adds collective mask to USER signals
+  typedef struct packed {
+    addr_t collective_mask;
+  } aw_user_t;
+
+  `AXI_TYPEDEF_AW_CHAN_T(mst_aw_chan_t, addr_t, id_mst_t, aw_user_t)
+  `AXI_TYPEDEF_AW_CHAN_T(slv_aw_chan_t, addr_t, id_slv_t, aw_user_t)
```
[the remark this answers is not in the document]
colluca: Right, I see what you mean now. Good catch, `aw_user_t` should be a type parameter of the module.

src/axi_mcast_demux_mapped.sv
```diff
@@ -0,0 +1,547 @@
...
+  ) i_aw_spill_reg (
+    .clk_i,
+    .rst_ni,
+    .valid_i ( slv_req_i.aw_valid    ),
+    .ready_o ( slv_resp_o.aw_ready   ),
+    .data_i  ( slv_req_i.aw          ),
+    .valid_o ( slv_req_cut.aw_valid  ),
+    .ready_i ( slv_resp_cut.aw_ready ),
+    .data_o  ( slv_req_cut.aw        )
+  );
```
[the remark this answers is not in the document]
colluca: I understand your concern.

The `axi_demux` still preserves the previous behaviour. Among the old modules, this change only affects the `axi_xbar` and the `axi_xbar_unmuxed`, because I based the latter on the `axi_mcast_xbar_unmuxed`, to improve code reuse and maintainability.

However, I have nothing against keeping the old implementations of the `axi_xbar` and `axi_xbar_unmuxed`. This should solve the problem you mention.

It's just subject to this tradeoff, so I would leave the decision to you.
Anyways, what is truely important for maintainability is that the building blocks (i.e. `demux_simple` and `mux`) wrap the multicast variants; for these other modules I think it's not important.

Let me know what you think and I'll revert the `axi_xbar` and `axi_xbar_unmuxed` if needed.

colluca (commented):
I've separated all orthogonal changes to individual PRs as requested.

colluca:
> Thanks. Regarding the fundamental question of whether we should keep multicast features separate from the core AXI code, have we reached a conclusion yet?

I think the idea is to proceed with the current implementation, or a small variant of it. That is, the unicast IPs would wrap the multicast IPs, when this is possible in a backwards-compatible manner, i.e. when by configuring the multicast IPs appropriately the functionality degenerates to the previous unicast functionality.

The only remaining violation to this is in the `axi_xbar` and `axi_xbar_unmuxed`, where the only difference between the previous and current implementation (based on the multicast variants) is the location of the spill registers, after and before the address decoders, respectively. This can easily be addressed, see the discussion above for more details.

Bender.yml
imchenwu, line 1: Check again the file level. 

`axi_mcast_mux` instantiates `axi_id_prepend` so it should be moved to Level3. Also `axi_mcast_demux_mapped` instantiates `axi_err_slv` so it should be moved to Level4, and all modules hierarchically instantiating this module should also be moved to correct level.

src/axi_pkg.sv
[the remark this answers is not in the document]
imchenwu: I agree with @micprog and I prefer to keep the original `xbar_cfg_t` and set `NoMulticastRules` and `NoMulticastPorts` as the module parameters.

src/axi_mcast_mux.sv
```diff
@@ -0,0 +1,672 @@
...
+  input  mst_resp_t                  mst_resp_i
+);
+
+  localparam int unsigned MstIdxBits    = $clog2(NoSlvPorts);
+  localparam int unsigned MstAxiIDWidth = SlvAxiIDWidth + MstIdxBits;
+
+  // pass through if only one slave port
+  if (NoSlvPorts == 32'h1) begin : gen_no_mux
+    cc_spill_register #(
+      .data_t   ( mst_aw_chan_t ),
```
imchenwu, line 75: check the small indentation issues in thie PR.

src/axi_mcast_mux.sv
```diff
@@ -0,0 +1,665 @@
...
+              lock_aw_valid_d = 1'b1;
+              load_aw_lock    = 1'b1;
+              // remember if it was a unicast transaction
+              lock_unicast_d  = !mcast_aw_commit;
+            end else begin
+              w_fifo_push = 1'b1;
+            end
+          end
+        end
+      end
```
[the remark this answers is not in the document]
imchenwu: See my comments in #408.

README.md
imchenwu, line 1: This `README.md` should be updated.

src/axi_mcast_demux_mapped.sv
```diff
@@ -0,0 +1,547 @@
...
+
+  `AXI_ASSIGN_TO_REQ(slv_req, slv)
+  `AXI_ASSIGN_FROM_RESP(slv, slv_resp)
+
+  for (genvar i = 0; i < NO_MST_PORTS; i++) begin : gen_assign_mst_ports
+    `AXI_ASSIGN_FROM_REQ(mst[i], mst_req[i])
+    `AXI_ASSIGN_TO_RESP(mst_resp[i], mst[i])
+  end
+
+  axi_mcast_demux #(
```
imchenwu, line 510: This module does not exist now. Do you mean `axi_mcast_demux_mapped` here?

src/axi_mcast_xbar_unmuxed.sv
```diff
@@ -0,0 +1,286 @@
...
+  parameter bit [Cfg.NoSlvPorts-1:0][Cfg.NoMstPorts-1:0] MULTICAST_CONNECTIVITY = '1,
+  parameter type rule_t                 = axi_pkg::xbar_rule_64_t
+) (
+  input  logic                                                      clk_i,
+  input  logic                                                      rst_ni,
+  AXI_BUS.Slave                                                     slv_ports [Cfg.NoSlvPorts-1:0],
+  AXI_BUS.Master                                                    mst_ports [Cfg.NoMstPorts-1:0][Cfg.NoSlvPorts-1:0],
+  input  rule_t [Cfg.NoAddrRules-1:0]                               addr_map_i,
+  input  logic  [Cfg.NoSlvPorts-1:0]                                en_default_mst_port_i,
+  input  logic  [Cfg.NoSlvPorts-1:0][idx_width(Cfg.NoMstPorts)-1:0] default_mst_port_i
```
imchenwu, line 225: The type of the internal `default_mst_port_i` of `axi_mcast_xbar_unmuxed` is `rule_t [NoSlvPorts-1:0]`.

src/axi_mcast_demux_mapped.sv
```diff
@@ -0,0 +1,547 @@
...
+  parameter bit                  SpillB                = 1'b0,
+  parameter bit                  SpillAr               = 1'b1,
+  parameter bit                  SpillR                = 1'b0,
+  parameter bit [NoMstPorts-1:0] Connectivity          = '1,
+  parameter bit [NoMstPorts-1:0] MulticastConnectivity = '1,
+  parameter type                 rule_t                = logic,
+  parameter int unsigned         NoAddrRules           = 32'd0,
+  parameter int unsigned         NoMulticastRules      = 32'd0,
+  parameter int unsigned         NoMulticastPorts      = 32'd0,
+  parameter int unsigned         MaxMcastTrans         = 32'd7
```
imchenwu, line 69: Is this parameter configurable for the upstream modules? I don't see any xbar modules passing this parameter.

test/axi_synth_bench.sv
```diff
@@ -793,3 +787,1011 @@ module synth_axi_lite_dw_converter #(
   );
 
 endmodule
+
+
+module synth_axi_xbar #(
```
imchenwu, line 792: You didn't instantiate this module in the top `axi_synth_bench` module so it was never tested in the ci flow..

src/axi_mcast_xbar.sv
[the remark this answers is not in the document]
imchenwu: It looks like the original `aw_user_t` is hardcoded inside the module and now it is a parameter default set according to `AXI_USER_WIDTH`. I have some concerns:

1. The default `aw_user_t` cannot be used when `Cfg.NoMulticastPorts > 0`: `axi_mcast_demux_mapped` accesses `aw.user.collective_mask`, which a plain vector does not have, so elaboration will fail. 
2. Below the line 222: `AXI_USER_WIDTH` still only controls the interface and the W/B/AR/R channels; the AW user width is `$bits(aw_user_t)`. When the two differ the macro in line 222 silently truncates or zero-extends between the interface vector and the struct..

What about only keeping `parameter type user_t` and getting `AXI_USER_WIDTH` as `localparam = $bits(user_t)`? 

Note: same issue occur in the other mcast `_intf` wrappers. (It's always a bit tricky at these `_intf` modules because ci doesn't really cover these interface modules...)

src/axi_mcast_demux_mapped.sv
```diff
@@ -0,0 +1,547 @@
...
+          dec_aw_select_mask = {'0, dec_aw_multicast_select_mask} & MulticastConnectivity;
+        end
+      end
+    end
+
+  end else begin : gen_no_multicast
+    assign dec_aw_addr = {'0, {NoMstPorts{slv_req_cut.aw.addr}}};
+    assign dec_aw_mask = '0;
+    assign dec_aw_select_mask = (dec_aw_unicast_error) ? select_error_slave :
+                                (dec_aw_unicast_select_mask & Connectivity);
```
imchenwu, line 328: A possible case here: addr decoding is legal but not connected (Connectivity is 0), then this `dec_aw_select_mask = 0`. Then inside `axi_mcast_demux_simple` module, the `accept_aw` signal is set to 1, because all-0 select signal is treated as "all ports are ready" to accept.. I kind of understand that the filter here is designed for multicast, but it breaks a case in unicast in my understanding.

src/axi_mcast_demux_simple.sv
```diff
@@ -0,0 +1,746 @@
...
+  typedef logic [IdCounterWidth-1:0] id_cnt_t;
+
+  localparam int unsigned McastCounterWidth = (MaxMcastTrans > 32'd1) ? $clog2(MaxMcastTrans+1) : 32'd1;
+  typedef logic [McastCounterWidth-1:0] mcast_cnt_t;
+
+  // pass through if only one master port
+  if (NoMstPorts == 32'h1) begin : gen_no_demux
+    `AXI_ASSIGN_REQ_STRUCT(mst_reqs_o[0], slv_req_i)
+    `AXI_ASSIGN_RESP_STRUCT(slv_resp_o, mst_resps_i[0])
+  end else begin : gen_demux
```
imchenwu, line 87: Multicast and unicast are coupled in this block, so even a unicast configuration (`NoMulticastPorts == 0`, i.e. every existing `axi_demux` / `axi_xbar` user) now pays for logic that only exists for multicast: the popcount and `aw_is_multicast`, the outstanding-multicast counter and everything it gates on the B channel (stream_join, lzc, per-port resp merge), the AW handshake FSM, and the one-hot/binary round trip of the select. Most of this is state-dependent and will not be optimized away.

Could the multicast-specific parts be wrapped in a generate block on `NoMulticastPorts > 0`, the way `UniqueIds` and `AtopSupport` are handled, so that the unicast elaboration stays equivalent to master? Since `UniqueIds` and multicast are already mutually exclusive (L710), the ID-counter path would not even need a multicast variant, which should keep the split straightforward. What do you think?

imchenwu (changes requested):
I think there are still some issues to be cleaned up, see the comments below. Also this multicast feature needs a documentation to tell users how a multicast transaction should be like (e.g. how is the user field of a mcast req like), and the current limitations of multicast feature. Optimally a block diagram for clear explanation would be even better.

PR #399, opened 2025-10-01
src: axi_fifo_delay_dyn: Fix incorrect power-of-two check for `Depth` parameter

yanjiuntw:
The `axi_fifo_delay_dyn` module's parameter constraint checker incorrectly identified valid `Depth` values as not being a power of two.

This commit corrects the logic to accurately verify that the `Depth` parameter is a power of two.

### DIFF
--- a/src/axi_fifo_delay_dyn.sv
+++ b/src/axi_fifo_delay_dyn.sv
@@ -283,7 +283,7 @@ module stream_fifo_delay_dyn #(
   `endif
   `endif
 
-  if (Depth & (Depth - 1) == 0)
+  if ((Depth & (Depth - 1)) != 0)
     $fatal(1, "Depth must be a power of two");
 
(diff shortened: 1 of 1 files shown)


src/axi_fifo_delay_dyn.sv
```diff
@@ -283,7 +283,7 @@ module stream_fifo_delay_dyn #(
   `endif
   `endif
 
-  if (Depth & (Depth - 1) == 0)
+  if ((Depth & (Depth - 1)) != 0)
```
imchenwu:
```suggestion
  if (Depth == 0 || (Depth & (Depth - 1)) != 0)
```

imchenwu (changes requested):
thx for contribution.

imchenwu (approved):

PR #407, opened 2026-01-29
`axi_xbar`: Remove redundant assertions

colluca:
The same assertions can be found within `axi_xbar_unmuxed`:
https://github.com/pulp-platform/axi/blob/814b00076cb4439c541191f793d6630e1d219bd5/src/axi_xbar_unmuxed.sv#L256-L267
So these are redundant:
 https://github.com/pulp-platform/axi/blob/814b00076cb4439c541191f793d6630e1d219bd5/src/axi_xbar.sv#L157-L168

### DIFF
--- a/src/axi_xbar.sv
+++ b/src/axi_xbar.sv
@@ -154,18 +154,6 @@ import cf_math_pkg::idx_width;
     );
   end
 
-  // pragma translate_off
-  `ifndef VERILATOR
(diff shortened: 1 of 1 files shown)


micprog (approved):
As the axi_xbar instantiates the axi_xbar_unmuxed which already has the given assertions, I think this minimal change is OK.

PR #408, opened 2026-01-29
`axi_mux`: Remove dependency of ready on valid

colluca:
This change was originally introduced in https://github.com/pulp-platform/axi/pull/398, though it may no longer be strictly required.

However, when this is possible, it is in general good to remove dependencies of `aw_ready` on `aw_valid`, as the dependency creates paths that are double in length, for de-asserting valid (from subordinate->manager to manager->subordinate->manager).

This PR further changes the condition which determines when the AW channel (or the arbitration outcome carried within it) is pushed into the W FIFO.
Instead of pushing it as soon as there is a valid AW, it only pushes when the AW is handshaked.
I don't know if there was a specific reason for this to be implemented as it was, but from what I understand this change shouldn't have any relevant performance implications, and is in my opinion safer.

Probably we could also decouple ready and valid, while reverting the `w_fifo_push` change, if needed. But, it would remain to be tested whether this would work with multicast.

### DIFF
--- a/src/axi_mux.sv
+++ b/src/axi_mux.sv
@@ -296,17 +296,23 @@ module axi_mux #(
           aw_ready        = 1'b1;
           lock_aw_valid_d = 1'b0;
           load_aw_lock    = 1'b1;
+          w_fifo_push     = 1'b1;
         end
       end else begin
-        if (!w_fifo_full && aw_valid) begin
-          mst_aw_valid = 1'b1;
-          w_fifo_push = 1'b1;
+        if (!w_fifo_full) begin
           if (mst_aw_ready) begin
             aw_ready = 1'b1;
-          end else begin
-            // go to lock if transaction not in this cycle
-            lock_aw_valid_d = 1'b1;
-            load_aw_lock    = 1'b1;
+          end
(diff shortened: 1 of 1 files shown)


imchenwu (commented):
According to my understanding, this PR mainly works on two things:

1. Remove dependency of aw_ready on aw_valid.
2. Originally W is pushed into the fifo when AW is valid, and now W is pushed when the handshake of AW is successful.

Regarding the first point, it looks good to me, although ready is, strictly, still dependent on valid because of the `rr_arb_tree` logic.

Regarding the second point, it exactly safer in some sense, but I am really concerned about a deadlock situation. The AXI spec allows a slave to assert aw_ready only when it receives a valid W, but here we only send a valid W when AW handshake is successful (aw_ready & aw_valid). [axi_to_detailed_mem](https://github.com/pulp-platform/axi/blob/4da15979747f326bde2f9869c64e587ce599772c/src/axi_to_detailed_mem.sv#L239) is exactly such kind of slave... I prefer to revert the `w_fifo_push` change, and if there's a problem in multicast, we can discuss another solution then.

PR #409, opened 2026-02-05
id_remap: enable non-atop writes when reads are stalled and vice versa

mrogenmoserTT:


src/axi_id_remap.sv
```diff
@@ -302,23 +302,93 @@ module axi_id_remap #(
...
+            state_d = HoldAR;
+          end else begin
+            state_d = Ready;
+          end
+        end else begin
+          if (mst_req_o.ar_valid && !mst_resp_i.ar_ready) begin
+            state_d = HoldAx;
+          end else begin
+            state_d = HoldAW;
+            ar_prio_d = 1'b0; // Reset AR priority, because no handshake was successful in this cycle.
```
imchenwu: I'm confused why you reset AR priority here. Shouldn't it only be reset when an AR's handshake is successful?

mrogenmoserTT: It's a bit confusing to me too. I adjusted the `ar_prio` scheme to only give AR priority if it is actually waiting, thus removing the obligation to clear here.

imchenwu (approved):

PR #413, opened 2026-02-23
Add `DECL` macros to be used in parameter port list

ricted98:
We can split the `TYPDEF` macro into `TYPEDEF` (backward compatible to define a new type) and `DECL` (to be used for `type` parameters), as done [here](https://github.com/openhwgroup/cv-hpdcache/blob/master/rtl/include/hpdcache_typedef.svh). These macros are particularly useful e.g. for synthesis wrappers or for sane type parameters defaults.


micprog (approved):
LGTM :+1:

thommythomaso:
LGTM

PR #414, opened 2026-02-25
Add github action for verilator lint and yosys-slang elaboration

micprog:


src/axi_test.sv
```diff
@@ -2145,7 +2145,7 @@ package axi_test;
                             Exp Data: %p \n \
                             Act Data: %h \n \
                             BeatData: %h",
-                  r_beat.r_id, beat_address+j, idx_data, exp_data, act_data, r_beat.r_data);
+                  r_beat.r_id, beat_address+j, idx_data, exp_data[0], act_data, r_beat.r_data);
```
imchenwu: may I know the intention of this line?

Bender.yml
```diff
@@ -19,7 +19,7 @@ package:
     - "Florian Zaruba <zarubaf@iis.ee.ethz.ch>"
 
 dependencies:
-  common_cells: { git: "https://github.com/pulp-platform/common_cells.git", version: 1.37.0 }
+  common_cells: { git: "https://github.com/pulp-platform/common_cells.git", version: 1.39.0 }
   common_verification: { git: "https://github.com/pulp-platform/common_verification.git", version: 0.2.5 }
   tech_cells_generic:  { git: "https://github.com/pulp-platform/tech_cells_generic.git", version: 0.2.2 }
```
imchenwu, line 24: maybe align the version of `tech_cells_generic` with `Bender.lock`?

src/axi_test.sv
micprog: `exp_data` is an infinite queue. We only want to print element 0, the front of this queue. Some simulators automatically infer this, but the old version is not correct systemverilog when printing into a `%h` specifier.

Bender.yml
micprog: Generally, these are minimum requirements. For common_cells, we are using a module that is only available v1.39.0 onwards. The tech_cells requirement does not need to precisely align with the lockfile, only the minimum we need, and afaik this minimum is OK as it is compatible with the interface we are specifying and the features we are using. It may be reasonable to update, as there were a few fixes, but AFAIK tc_sram is only used for https://github.com/pulp-platform/axi/blob/master/test/tb_axi_to_mem_banked.sv#L143-L160 .

micprog: Arguably this dependency could be guarded behind the test target using new the new bender features.

src/axi_sim_mem.sv
```diff
@@ -341,10 +341,10 @@ module axi_sim_mem #(
...
   // Parameter Assertions
   initial begin
-    assert (AddrWidth != 0) else $fatal("AddrWidth must be non-zero!", 1);
-    assert (DataWidth != 0) else $fatal("DataWidth must be non-zero!", 1);
-    assert (IdWidth != 0) else $fatal("IdWidth must be non-zero!", 1);
-    assert (UserWidth != 0) else $fatal("UserWidth must be non-zero!", 1);
+    assert (AddrWidth != 0) else $fatal(1, "AddrWidth must be non-zero!", 1);
+    assert (DataWidth != 0) else $fatal(1, "DataWidth must be non-zero!", 1);
+    assert (IdWidth != 0) else $fatal(1, "IdWidth must be non-zero!", 1);
+    assert (UserWidth != 0) else $fatal(1, "UserWidth must be non-zero!", 1);
```
imchenwu:
```suggestion
    assert (AddrWidth != 0) else $fatal(1, "AddrWidth must be non-zero!");
    assert (DataWidth != 0) else $fatal(1, "DataWidth must be non-zero!");
    assert (IdWidth != 0) else $fatal(1, "IdWidth must be non-zero!");
    assert (UserWidth != 0) else $fatal(1, "UserWidth must be non-zero!");
```

imchenwu (changes requested):

src/axi_test.sv
micprog: You are correct, my bad. This was fixed here this way, but better using `%p` on master, so this change is no longer needed. I reverted the commit in question.

imchenwu (approved):

nikgiu (approved):

PR #417, opened 2026-03-31
tb_axi_xbar: Fix queue format specifier

niwis:
`exp_data` is a queue. To print its contents, the `%p` format specifier for unpacked aggregate types needs to be used instead of `%h`. Fixes a slang compilation error.


imchenwu (approved):
LGTM

PR #418, opened 2026-04-10
Fix user/id width in axi_to_detailed_mem_intf

gtkiku:
I happened to notice that `axi_to_detailed_mem_intf` currently has each array element of `mem_user_o` and `mem_id_o` as a single bit wide, where they presumably should be `USER_WIDTH` and `ID_WIDTH` wide, respectively.

The comments documenting the `axi_to_detailed_mem_intf` signals also refer to `axi_to_mem`, so I changed them to refer to `axi_to_detailed_mem` instead, thought I'd update them as well while I'm here.

I tried running the CI tests with `vsim 10.7c`, but even without these patches they fail with
```
# ** Error: AW burst crossing 4 KiB page boundary detected, which is illegal!
#    Time: 60 ns Started: 60 ns  Scope: tb_axi_bus_compare.axi_dv File: /home/kimi/Projects/tmp/axi/src/axi_intf.sv Line: 257
# ** Error: AW burst crossing 4 KiB page boundary detected, which is illegal!
#    Time: 70 ns Started: 70 ns  Scope: tb_axi_bus_compare.axi_dv File: /home/kimi/Projects/tmp/axi/src/axi_intf.sv Line: 257
# ** Note: $finish    : /home/kimi/Projects/tmp/axi/test/tb_axi_bus_compare.sv(307)
#    Time: 3360 ns  Iteration: 0  Instance: /tb_axi_bus_compare
# End time: 15:22:24 on Apr 10,2026, Elapsed time: 0:00:01
```
so I'm curious to see what the actual CI reports. I've checked that the modifications work for my use case at least.

### DIFF
--- a/src/axi_to_detailed_mem.sv
+++ b/src/axi_to_detailed_mem.sv
@@ -623,7 +623,11 @@ module axi_to_detailed_mem_intf #(
   /// Dependent parameter, do not override. See `axi_to_mem`, parameter `mem_data_t`.
   localparam type mem_data_t = logic [DATA_WIDTH/NUM_BANKS-1:0],
   /// Dependent parameter, do not override. See `axi_to_mem`, parameter `mem_strb_t`.
-  localparam type mem_strb_t = logic [DATA_WIDTH/NUM_BANKS/8-1:0]
+  localparam type mem_strb_t = logic [DATA_WIDTH/NUM_BANKS/8-1:0],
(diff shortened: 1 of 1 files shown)


imchenwu:
Looks good to me. Internal CI uses vsim 2025.1 and seems like tests have passed.

imchenwu (approved):

PR #424, opened 2026-06-15
ci: replace memora with Gitlab-native artifacts and rules

fischeti:
Drop the custom [memora](https://github.com/andreaskurth/memora-rs) caching
tool in favor of GitLab-native mechanisms:

- The compiled vsim `work` library is now passed from the build to the test
  stage as a regular GitLab artifact instead of a memora cache.
- Per-module `rules:changes` lists, derived from the old `Memora.yml` inputs,
  skip simulation jobs unless one of their source dependencies changed,
  recovering memora's selective-rerun behavior natively.
- Every `rules:changes` rule compares against `refs/heads/master`
  (`compare_to`) rather than the previous commit. This makes a trigger reflect
  the cumulative diff of the whole branch, like memora's content-addressed
  cache did: a file touched anywhere on the branch keeps triggering its job
  until merged. Without this, a later unrelated commit would re-run the
  pipeline without the earlier change in its `changes` diff, hiding a failing
  job behind a green pipeline.
- The per-module simulation jobs now share a single `.run_vsim` template. Leaf
  modules are grouped in a matrix job, while modules with shared submodule
  dependencies use the same common rule structure plus a local dependency list.

### Caveats

- The dependency lists are hand-maintained, exactly as the old `Memora.yml`
  `inputs` were; a missing entry silently skips a test. Only `Bender.yml`
  itself is tracked, so a change to an *external* dependency still triggers
  every job (as before).
- Memora cached test *results* by content hash, so a push that left a module's
  inputs unchanged was a cache hit and skipped re-running it. `rules:changes`
  is not a cache: a module that differs from `master` re-runs on every push to
  the branch until merged, even if its files did not change since the last
  pipeline. This does not reduce coverage, but it triggers jobs more often than
  memora did.

### Additional cleanups

- Collapse the single-entry `VSIM_VER` matrix and promote `VSIM` to a global
  variable; drop the unused `VLOG`/`VLIB` exports.
- Use GitLab artifacts for the compiled simulation library instead of custom
  cache lookup/insert scripts.
- Use `uvx` instead of `pip install` for FuseSoC in `fuse_xsim`.
- Bump the GitHub Actions checkout and Bender installer actions.
- Restore documentation deployment for version tags while keeping deployment
  restricted to upstream `master` pushes and `v*` tag pushes.


.gitlab-ci.yml
```diff
@@ -1,190 +1,257 @@
...
-        mv work{-$VSIM_VER,}
-        ../scripts/run_vsim.sh --random-seed $TEST_MODULE && touch $ARTIFACT.tested
-        $CI_PROJECT_DIR/.gitlab-ci.d/memora_retry.sh insert $ARTIFACT
-      fi
+    - cd build && ../scripts/run_vsim.sh --random-seed $TEST_MODULE
+  rules:
+    - &run_vsim_common_change_rule
+      changes:
+        compare_to: 'refs/heads/master'
+        paths:
```
imchenwu, line 86: I think it makes sense to run vsim for all modules whenever the vsim version changes, as the original memora-based CI did.
```suggestion
        paths:
          - .gitlab-ci.yml
```

imchenwu (commented):
LGTM except for the small point as below.

.gitlab-ci.yml
fischeti: makes sense! I added it back

imchenwu (approved):
Wait for review from @nikgiu

nikgiu (approved):

Issue #353, opened 2024-10-09
axi_xbar_unmuxed: Avoid multi dimensional arrays in more tools


olofk:
axi_xbar_unmuxed doesn't parse with (our version of) Xcelium because of problems with multi dimensional interfaces. I see the same problem has come up with VCS which was fixed in https://github.com/pulp-platform/axi/commit/14fbbf59082c6b13e8820a71131b8f845e3a8683

Two thoughts here
1. My experience tells me that it's best to avoid multi-dimensional interfaces completely because of poor tool support
2. Use something like `` `ifdef NO_MULTI_DIM_ARRAYS`` instead of `` `ifdef VCS``. It's not a good idea to have tool-specific ifdefs. It's better to name the ifdef after the actual missing feature (e.g. `` `ifdef NO_MULTI_DIM_ARRAYS`` in this case) I had the same comment a couple of years ago around assertions where there were a lot of `` `ifdef VERILATOR``. When we wanted to run this in XSIM, we had to also add a layer of `` `ifdef XSIM`` and then after a while, Verilator started to support assertions which made things more confusing. Having e.g. an `` `ifdef ASSERTIONS`` and controlling when to set that through the build system would help in these cases.

olofk:
Just remembered that common_cells nowadays does have a `` `COMMON_CELLS_ASSERTS_OFF`` which I think is a good example of how to do this.

foreverleo:
Yes, I agree. I am currently using Synopsys's Spyglass for Lint checking. The ``` `ifndef VCS``` got bypassed and caused an error.

PR #428, opened 2026-06-24
axi_xbar_unmuxed: remove multi-dimensional interface.

nikgiu:
Use ifdef QUESTA for 2-D mst_ports, flat elsewhere

PR #350 and #405 worked around poor multi-dimensional interface array support by
wrapping the entire axi_xbar_unmuxed_intf module in tool-specific ifndef guards,
effectively hiding it from VCS and Genus. However, this excludes the module from an
ever-growing list of tools (see issue #353).

Instead, scope the guard to just the mst_ports port declaration and its assign
loop: under QUESTA, the original 2-D array mst_ports[i][j] is preserved for
readability in waveform viewers; all other tools get a flat row-major 1-D array
mst_ports[i*Cfg.NoSlvPorts + j].


src/axi_xbar_unmuxed.sv
```diff
@@ -286,7 +283,14 @@ import cf_math_pkg::idx_width;
   input  logic                                                      rst_ni,
   input  logic                                                      test_i,
   AXI_BUS.Slave                                                     slv_ports [Cfg.NoSlvPorts-1:0],
+`ifdef QUESTA
```
imchenwu: Where is this "QUESTA" macro defined? Is this supposed to be passed manually? AFAIK, we have a macro that Bender automatically defines when targeting VSIM.
```suggestion
`ifdef TARGET_VSIM
```

imchenwu (changes requested):
Besides the VSIM macro naming, another concern I have is that this intf module isn't really used often. Can we just guard this intf module with the TARGET_VSIM macro and don't compile/use it at all for all other tools?

src/axi_xbar_unmuxed.sv
nikgiu: I'll switch to TARGET_VSIM for consistency with the rest of the repo. Bender defines it automatically when targeting vsim, so it's the more reliable choice here. QUESTA would also work in principle since Questa sets it as a built-in predefined macro, but TARGET_VSIM is the correct convention to use in this codebase.

nikgiu:
Looking at other _intf modules in the repo (axi_cut_intf, axi_isolate_intf, axi_fifo_intf, etc.), none of them are wrapped in tool-specific guards.They all compile unconditionally. Wrapping axi_xbar_unmuxed_intf in ifdef TARGET_VSIM would be the only exception to this pattern and would silently deny access to the module for users on VCS, Xcelium, or any other tool.

That said, the old ifndef VCS / ifndef TARGET_GENUS approach was already an exception to this convention, so I understand the argument.

imchenwu:
I still feel it is inelegant to switch the port geometry on a define, which means every downstream instantiation has to replicate the same ifdef to connect mst_ports, so the tool dependence leaks to all users instead of being encapsulated. It also gives us two APIs to maintain within a single module.

I would still prefer to delete this intf module because it's not used much anyway and is not even instantiated or tested in this repo. There are still some other modules that don't have their corresponding intf module, so I think deleting this intf module should be fine.

PR #430, opened 2026-06-27
docs: clarify data-width converters' burst-type support (INCR/WRAP/FI…

### LINKED ISSUE #429: Docs: clarify the DW converters' burst-type support (WRAP / FIXED) — documentation gap, not a bug
> **This is a documentation-clarification issue, not a bug report.** The runtime behaviour is intentional and correct; only the public documentation is silent about it.

**Affected:** `axi` live HEAD `e55ae2a` — `src/axi_dw_downsizer.sv`, `src/axi_dw_converter.sv`, and the top-level `README.md` module table.

## The gap

The data-width converters have real, intentional burst-type restrictions, but they are only discoverable by reading the SystemVerilog source headers — the public docs don't mention them, so a user first learns of them by getting an unexpected `SLVERR` (or by seeing `FIXED` silently become `INCR` downstream). Concretely: (1) the `README.md` module table (the entry point) describes `axi_dw_converter` / `axi_dw_downsizer` / `axi_dw_upsizer` purely as "a data width converter…", with no mention of burst-type support; (2) there is no per-module doc page for the DW converters under `doc/` (unlike `axi_xbar`, `axi_mux`, etc.), so there is nowhere else to look; (3) the restriction *is* stated, but only in source headers, and even there incompletely — `axi_dw_downsizer.sv` (lines 18–21): WRAP → SLVERR, FIXED single-beat only, multi-beat FIXED → SLVERR; `axi_dw_converter.sv` (lines 14–16): "the downsizer also does not support FIXED bursts with incoming axlen != 0"; and neither header documents that a single-beat FIXED transaction requiring downsizing is re-encoded as `INCR` on the master (narrow) port.

## The undocumented behaviour (grounded in source)

`src/axi_dw_downsizer.sv` (read path ~435–463, write path ~521–547):

```systemverilog
axi_pkg::BURST_FIXED: begin
  if (r_req_d.ar.len == '0) begin            // single beat only
    // ... downsize ratio math ...
    if (conv_ratio != 1) begin
      r_req_d.ar.len   = r_req_d.burst_len  ;
      r_req_d.ar.burst = axi_pkg::BURST_INCR;   // <-- single-beat FIXED RE-ENCODED as INCR
    end
  end else begin
    r_req_d.ar_throw_error = 1'b1;           // multi-beat FIXED -> SLVERR
  end
end
axi_pkg::BURST_WRAP: begin
  r_req_d.ar_throw_error = 1'b1;             // WRAP -> SLVERR (unsupported)
end
```

So precisely: `INCR` is supported at any length; `WRAP` is not supported → `SLVERR`; `FIXED` is supported only single-beat (`axlen == 0`), and multi-beat `FIXED` → `SLVERR`. A single-beat wide `FIXED` that needs downsizing is re-encoded as `INCR` on the master port — address-equivalent for one beat, but a downstream observer sees `INCR`, not `FIXED`.

## Why this is NOT a bug

WRAP / multi-beat-FIXED → `SLVERR` is a deliberate, signalled rejection of an unsupported feature (not a silent miscompute). The single-beat FIXED → INCR re-encode is semantically correct: with exactly one beat the burst type has no observable address effect, so INCR and FIXED are equivalent at the byte level. The only gap is documentation — a user reading the README cannot discover any of this.

## Proposed fix

Add a burst-type-support note to the `README.md` module table covering all three DW modules: `INCR` any length; `WRAP` → `SLVERR`; `FIXED` single-beat only (`axlen == 0`), multi-beat `FIXED` → `SLVERR`; single-beat downsized `FIXED` re-encoded as `INCR` on the master port. (Optionally also append the re-encode note to the module headers so the source is self-consistent.) A ready README-only patch (clean `git apply --check` at HEAD `e55ae2a`, placed after the module table so it renders intact) is available — happy to open a PR.

## Verify

Documentation-only change — `git apply --check` clean at live HEAD `e55ae2a`; no RTL touched, no functional regression.

siju-felsite:
Closes #429.

Documentation-only: adds a note to the README module table making the data-width converters' burst-type support discoverable. Today it lives only in the SystemVerilog source headers, so a user first learns of it via an unexpected `SLVERR` or by seeing `FIXED` silently become `INCR` downstream.

The note covers all three DW modules (`axi_dw_converter` / `axi_dw_downsizer` / `axi_dw_upsizer`): `INCR` any length; `WRAP` -> `SLVERR`; `FIXED` single-beat only (`axlen == 0`), multi-beat `FIXED` -> `SLVERR`; a single-beat downsized `FIXED` is re-encoded as `INCR` on the master (narrow) port (address-equivalent for one beat, but a downstream observer sees `INCR`).

No RTL touched, no functional change. Placed after the module table so it renders intact; clean `git apply` at HEAD e55ae2a.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,8 @@ In addition to the documents linked in the following table, we are setting up [d
 | [`axi_xp`](src/axi_xp.sv)                                   | AXI Crosspoint (XP) with homomorphous slave and master ports.                                        |                                  |
 | [`axi_zero_mem`](src/axi_zero_mem.sv)                       | AXI-attached /dev/zero. All reads will be zero, writes are absorbed.                                 |                                  |
 
+> **Burst-type support of the data-width converters (`axi_dw_converter` / `axi_dw_downsizer` / `axi_dw_upsizer`).** These modules support `INCR` bursts of any length. `WRAP` bursts are **not** supported and are answered with `SLVERR`. `FIXED` bursts are supported **only when they consist of a single beat** (`axlen == 0`); a multi-beat `FIXED` burst is answered with `SLVERR`. Note that a single-beat `FIXED` transaction that requires downsizing is re-encoded as an `INCR` burst on the master (narrow) port — this is address-equivalent for one beat, but a downstream observer will see `INCR`, not `FIXED`.
+
 ## Synthesizable Verification Modules
 
 The following modules are meant to be used for verification purposes only but are synthesizable to be used in FPGA environments.


imchenwu:
Thx. I checked that the upsizer actually supports the `FIXED` type. I'll first close this PR and add the correct notes in the next release.

PR #431, opened 2026-07-02
feat: add native `axi_to_apb` converter

fischeti:
This PR adds a native `axi_to_apb` converter, which supports both downsizing datawidth and truncating the address.

## Current situation

It is already possible now to convert from AXI to APB but multiple modules are needed like 1) `axi_to_axi_lite ─▶ axi_lite_to_apb` or 2)  `axi_to_reg ─▶ reg_to_apb`. Both add intermediate protocol signals and generally make the code more verbose. Further, 2) uses a non-standard protocol and pulls in `register_interface` as a dependency.

Also, 1) has the disadvantage that it pulls in a `axi_to_axi_lite` which has relatively heavy burst splitter and other components, which is overkill to convert to APB. Further, neither `axi_to_axi_lite` nor `axi_lite_to_apb` allow to natively downsize the datawidth, hence a `axi_lite_dw_converter` is also needed.

## Exploration

I prototyped three ways to bridge AXI4+ATOP to APB4 and compared area (post-synth, gate equivalents):

| Approach | Area | |
|---|---|---|
| `axi_to_axi_lite` → `axi_lite_dw_converter` → `axi_lite_to_apb` chain | 11.6 kGE | |
| **`axi_to_detailed_mem` + inline APB master (this PR)** | **9.3 kGE** | ✅ chosen |
| Native AXI→APB FSM | 8.3 kGE | |

The AXI4-Lite chain was the heaviest: it drags in burst splitting, and a Lite data-width converter.

A fully native FSM was the smallest (~10% below the chosen option), but it was vibecoded so not proven logic, and code complexity is higher.

The chosen `axi_to_detailed_mem` + inline APB master sits in between: it reuses `axi_to_detailed_mem` the AXI conversion. It is ~20% smaller than the Lite chain.

## Verification

The adapter is verified with a `tb_axi_to_apb` testbench, that checks the APB interface for protocol compliancy, as well as an AXI scoreboard to check for correct data. I also added different types of slaves (ideal, random) to simulate backpressure.

### DIFF
--- a/test/tb_axi_to_apb.sv
+++ b/test/tb_axi_to_apb.sv
@@ -1,0 +1,404 @@
+// Copyright (c) 2024 ETH Zurich and University of Bologna.
+// Copyright and related rights are licensed under the Solderpad Hardware
+// License, Version 0.51 (the "License"); you may not use this file except in
+// compliance with the License.  You may obtain a copy of the License at
+// http://solderpad.org/licenses/SHL-0.51. Unless required by applicable law
+// or agreed to in writing, software, hardware and materials distributed under
+// this License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+//
+// Authors:
+// - Tim Fischer <fischeti@iis.ee.ethz.ch>
+
+// Description: Testbench for `axi_to_apb`.
+//
+// A full-AXI random master drives the DUT, whose APB master ports are attached to four behavioural
+// APB slaves with different timing: an ideal (0-wait), a slow (fixed-wait), a random-wait, and an
+// error slave. The first three are memory-backed, so `DUT + memory slaves` behaves as an AXI memory
+// and an `axi_scoreboard` on the AXI port checks end-to-end data integrity (address mapping, bank
+// placement, strobe masking, burst handling). The random master only targets the three memory
+// slaves; the error slave and an unmapped address are exercised by a few directed transactions that
+// check the error / decode-miss responses. APB-protocol legality is checked with assertions.
+
+`include "axi/typedef.svh"
+`include "axi/assign.svh"
+
+module tb_axi_to_apb #(
+  // Swept externally over {32, 64} to cover NumBanks = 1 and 2 (with ApbDataWidth = 32).
+  parameter int unsigned TbAxiDataWidth = 32'd64
+);
+
+  // Timing
+  localparam time CyclTime = 10ns;
+  localparam time ApplTime =  2ns;
+  localparam time TestTime =  8ns;
+
+  // Random master transaction counts
+  localparam int unsigned NoWrites = 2000;
+  localparam int unsigned NoReads  = 2000;
+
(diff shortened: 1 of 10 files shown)


test/tb_axi_to_apb.sv
```diff
@@ -0,0 +1,405 @@
...
+    .ApbDataWidth ( ApbDataWidth   ),
+    .axi_req_t    ( axi_req_t      ),
+    .axi_resp_t   ( axi_resp_t     ),
+    .apb_req_t    ( apb_req_t      ),
+    .apb_resp_t   ( apb_resp_t     ),
+    .rule_t       ( rule_t         )
+  ) i_dut (
+    .clk_i      ( clk      ),
+    .rst_ni     ( rst_n    ),
+    .test_i     ( 1'b0     ),
```
nikgiu:
```suggestion
```

the ci does not pass since is not present in the axi_to_abp module

nikgiu (changes requested):
While reviewing this I tried bumping OutFifoDepth to 2 in the testbench to see how the module behaves with more buffering, and the simulation blew up: Questa hits a zero-delay combinational loop (Autofindloop at 350 ns, X propagation, then vsimk SIGSEGV / exit 211). The loop it reports runs through mem_gnt/mem_rvalid in this module and back through the axi_to_detailed_mem response path (stream_join → stream_fork_dynamic).

Digging into it, I think the root cause is that bank_sel (and therefore apb_sel_idx) is purely combinational. It's the lzc output over mem_req, and it drives paddr/psel/pwdata/mem_gnt/… directly. So mem_gnt/mem_rvalid depend combinationally on mem_req, and once OutFifoDepth > 1 the front-end closes that path into a combinational cycle. With OutFifoDepth == 1 it happens to be safe only because a full depth-1 bank FIFO blocks the next beat's enqueue, so mem_req can't change mid-transfer.

I also think the same root cause can break the APB protocol even without hitting the loop: since the selection isn't frozen for the duration of a transfer, a lower-index bank asserting between SETUP and ACCESS shifts bank_sel/apb_sel_idx mid-transfer → SETUP is driven to one slave and ACCESS (penable) to another, and paddr/pwdata change while a transaction is in flight.

My suggestion would be to latch the selection at the SETUP→ACCESS transition and drive the bus from the latched values during ACCESS. That keeps the APB signals stable by construction and breaks the mem_req → bank_sel → mem_gnt combinational path, regardless of OutFifoDepth

fischeti:
Thanks for the detailed review @nikgiu! I will dig into the issue when I find some time again. I will draft it again, until I have fixed the issue.

fischeti:
You were right with your analysis @nikgiu! And I think your proposed fix is indeed the simplest option. I pushed some changes and also rebased to `devel` now.

nikgiu:
Thank you @fischeti ! Looks legit to me

nikgiu (approved):

imchenwu:
I noticed that in APB, `PSTRB` should be set to zero when reading, but in the previous code, the field `pstrb` was from the `mem_strb` from AXI W channel which is allowed to be non-zero during read transaction. I modified this to force `pstrb` to be 0.

fischeti:
Makes sense! Thanks

PR #439, opened 2026-07-29
axi_burst_splitter: add InOrderRsp burst bookkeeping

kgreigTT:
### What

Adds an `InOrderRsp` parameter to `axi_burst_splitter_gran` and `axi_burst_splitter`. When set, the burst keeper tracks outstanding split bursts with a plain FIFO of per-burst counters instead of the associative `id_queue` plus counter free list. That removes the ID hash tables, the per-slot `delta_counter` array and the free-slot `lzc`, which is a significant area saving in the configurations that can use it.

### Why it is safe to select

The FIFO path is only valid when downstream responses return in request order per channel. `axi_to_axi_lite` therefore hardwires it on, since an AXI4-Lite slave returns responses in request order per channel by protocol.

The parameter defaults to `1'b0`, so every existing instantiation keeps the current out-of-order `id_queue` path with no change in behaviour. The selection is purely structural: the two generate branches are interface- and handshake-identical, including the `FullBW=0` input-priority behaviour where the response port is not served in a cycle that takes an allocation.

### Verification

Both run with VCS on this branch, rebased on current `master` (includes #437):

- An equivalence testbench drove the in-order and legacy counter blocks in lockstep off identical stimulus and compared `alloc_gnt`, `cnt_len`, `cnt_err` and `cnt_gnt` every cycle, with randomised backpressure and error injection. All 6 configurations passed: `MaxTxns` 1/3/4 x `FullBW` 0/1, including `CutPath`.
- `tb_axi_to_axi_lite` passes with `InOrderRsp=1'b1`: 257577 write and 125378 read AXI4-Lite transactions, zero errors.

Opened as a draft. Happy to add the equivalence testbench to `test/` with its `Bender.yml` and `Makefile` registration if you would like it in-tree; it is currently kept outside this PR.

### Notes for review

- The one incidental change outside the new code is a width cast on the existing `idq_oup_pop` comparison (`cnt_len_o` is 8-bit, `cnt_delta_i` is 9-bit). It is behaviourally identical, since the narrower operand was already zero-extended, and it silences a lint warning. Happy to drop it if you would rather keep this PR to the feature.

### DIFF
--- a/src/axi_burst_splitter_gran.sv
+++ b/src/axi_burst_splitter_gran.sv
@@ -26,6 +26,8 @@ module axi_burst_splitter_gran #(
   parameter int unsigned MaxWriteTxns  = 32'd0,
   /// Internal ID queue can work in two bandwidth modes: see id_queue.sv for details
   parameter bit          FullBW        = 1'b0,
+  /// Downstream responses are guaranteed to arrive in request order
+  parameter bit          InOrderRsp    = 1'b0,
   /// Cut paths through the IP
   parameter bit          CutPath       = 1'b0,
   /// Disable checks, handle unsupported transfers as bypass instead of reporting an error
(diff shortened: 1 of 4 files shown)


src/axi_burst_splitter_gran.sv
```diff
@@ -447,16 +452,16 @@ module axi_burst_splitter_gran_ax_chan #(
   input  logic          ax_ready_i,
 
   input  id_t           cnt_id_i,
-  output axi_pkg::len_t cnt_len_o,
+  output axi_pkg::len_t  cnt_len_o,
```
imchenwu, line 455:
```suggestion
  output axi_pkg::len_t cnt_len_o,
```

imchenwu (commented):
Hi, thx for contribution! Overall I like this, but here we can reuse the non-fall-through `fifo_v3` module (see from https://github.com/pulp-platform/common_cells/blob/v1.39.0/src/fifo_v3.sv) when `InOrderRsp==1`. The only thing is that `fifo_v3` module doesn't support modifications to the entries, but as here we only modify the head entry, it should be okay to have a register to store the number of already consumed transactions.

Also note that [pulp-platform/common_cells](https://github.com/pulp-platform/common_cells) has released v2 (where `fifo_v3` is turned into `cc_fifo` module) and here axi will also have a new major release soon based on `devel` branch, so it would be great if you can rebase your work on the `devel` branch.

PR #440, opened 2026-08-02
axi_demux: enforce exactly MaxTrans in-flight transactions (fixes #249)

### LINKED ISSUE #249: axi crossbar MaxSlvTrans and MaxMstTrans not working as expected
Hello,

I noticed something strange abot the `MaxSlvTrans` and `MaxMstTrans` parameters of the axi crossbar.

First, I realized that they don't limit the number of transactions to exactly the parameter value, but is rather a lower bound. For example, adding registers (`LatencyMode = CUT_ALL_AX`) increase the number of transactions for all crossbar slaves. For the masters, there are FIFOs in the axi_mux and demux, which limit as expected the number of in flight transactions for the write channels. However, the reads are limited by the id counters present in the demux.

The issue I wanted to talk about is the width of these counters. In line 63 of `axi_demux.sv` we can see how the width is calculated:
```
localparam int unsigned IdCounterWidth = MaxTrans > 1 ? $clog2(MaxTrans) : 1;
```
The demux can send in flight transactions as long as the corresponfing id counter doesn't overflow (and for the case of writes, the FIFO is not full), which is usually more than the number specified in `MaxTrans`. However, when this value is a power of 2, the limit is reduced, and the maximum number of in flight transactions is `MaxTrans-1`, because the counter can't store the value `MaxTrans`.

For me it is not a problem that the crossbar can hold more in flight transactions that the specified number in `MaxSlvTrans` and `MaxMstTrans` (maybe it should be at least documented), but I it is an issue that it sends less than the specified number.
The solution I think is simple, just changing that line by:
```
localparam int unsigned IdCounterWidth = $clog2(MaxTrans+1);
```
The crossbar will still support more in flight transactions, but at least the parameters will set a lower bound.

What do you think on this? I am not an expert of this code so maybe I am leaving something out. Also, If you agree with the solution I can make a PR.

Thanks!

### LINKED ISSUE #336: "MaxTrans" defined incorrectly in the module documentation
https://github.com/pulp-platform/axi/blame/ac5deb3ff086aa34b168f392c051e92603d6c0e2/doc/axi_demux.md#L27

MaxTrans:  **For each ID**, the slave port can have at most this many transactions in flight

arjunsv98:
## Summary

The demux's per-ID in-flight counters are sized with `idx_width(MaxTrans)` bits — enough to *index* `MaxTrans` values (0..MaxTrans-1), not enough to *count* to `MaxTrans`. Consequently:

- **For power-of-two `MaxTrans` — including the module default of 8** — the counters report full at `MaxTrans - 1`: a demux configured for 8 outstanding transactions accepts only 7, on the AR path, the AW path, and the open-W-burst (`w_open`) ceiling.
- For other values, the all-ones full condition allows *more* than configured (e.g. 15 in flight for `MaxTrans = 10`).

This is exactly the behavior diagnosed by the reporter of #249 in 2022; this PR implements the fix (with exact-limit semantics) and adds the regression that was missing.

## Fix

- `axi_demux_simple`: widen `IdCounterWidth` to `idx_width(MaxTrans + 1)`; gate `w_open` with `< MaxTrans` instead of `!= all-ones`; pass `MaxTrans` to the ID counters.
- `axi_demux_id_counters`: new `MaxTrans` parameter (defaults to the legacy saturation value, so standalone instantiations are unaffected); full condition compares `in_flight >= MaxTrans` instead of `&in_flight`. `>=` rather than `==` because a simultaneous push + ATOP injection advances a counter by 2.

**Note on semantics:** this makes `MaxTrans` an exact limit. Non-power-of-two configurations previously enjoyed slack up to the counter's all-ones value (e.g. 15 for `MaxTrans=10`); they are now limited to exactly `MaxTrans`. This matches the parameter's documentation (see also #336).

## Verification

New directed test `tb_axi_demux_maxtrans`: back-to-back same-ID reads into a demux whose selected master port accepts every AR but withholds all R responses, so the in-flight count can only grow; the number of accepted AR handshakes equals the true counter limit.

| config | before | after |
|---|---|---|
| `MaxTrans = 8` | **7** accepted (bug) | **8** |
| `MaxTrans = 10` | 15 accepted (slack) | **10** |

Full `tb_axi_xbar` regression (Verilator): 8072 checks, 0 failures, in both the default (10/6) and an all-power-of-two (8/8) `MaxMstTrans`/`MaxSlvTrans` configuration.

No existing test could observe this bug: the shipped testbenches all use non-power-of-two limits, and none measures achieved concurrency — the new directed test closes that gap.

Fixes #249


test/tb_axi_demux_maxtrans.sv
```diff
@@ -0,0 +1,153 @@
...
+// reports "full" one transaction early.
+//
+// Method: issue back-to-back same-ID reads into a demux whose selected master
+// port accepts every AR but never returns R beats, so the in-flight count can
+// only grow.  The number of accepted AR handshakes before `ar_ready`
+// deasserts is exactly the counter limit.  Expected: `MaxTrans`.
+
+`include "axi/typedef.svh"
+
+module tb_axi_demux_maxtrans #(
```
imchenwu: Is this tb necessary? I suggest to add assertions inside the `axi_demux_id_counters` module instead.

src/axi_demux_id_counters.sv
```diff
@@ -125,7 +129,9 @@ module axi_demux_id_counters #(
       .overflow_o ( overflow  )
     );
     assign occupied[i] = |in_flight;
```
imchenwu: A small issue here is that when there are `2^WIDTH` outstanding transactions, the `any_outstanding_trx_o` will report 0 outstanding. 
```suggestion
    assign occupied[i] = |{overflow, in_flight};
```

imchenwu (changes requested):
Hi, thx for working on this. :) As this PR requires exposing a new port, we'll have to merge it in the next major release, so I modified the base branch to `devel`.

arjunsv98:
Thanks for taking a look! Applied your `occupied` suggestion and rebased onto `devel`.

About the tb — I considered moving the check into assertions in `axi_demux_id_counters` instead, but realized they'd never actually trigger in CI: all the existing testbenches use non-power-of-2 `MaxTrans` values (10/6/5), and the bug only manifests for power-of-2 configs. So an assertion would protect downstream users, but the repo's own CI would still have passed with the original bug in place. The directed tb is what actually builds a triggering config.

Happy to add the assertions on top if you'd like both, or drop the tb if you feel strongly — just wanted to flag why I kept it for now.

test/tb_axi_demux_maxtrans.sv
arjunsv98: Replied in the main thread with more detail, but short version: the existing tb configs all use non-power-of-2 MaxTrans, so module assertions alone would never fire in CI — this tb is what actually builds a triggering config. Can add assertions on top or drop the tb, whichever you prefer.

src/axi_demux_id_counters.sv
arjunsv98: Good catch, applied.

imchenwu:
Hi, I checked the current CI. Actually `tb_axi_dw_upsizer` sets `AXI_MAX_READS` as 4, which reaches `axi_demux` inside `axi_dw_upsizer` as `MaxTrans=4`. Therefore, I still prefer to use the assertion instead of this specialized tb.

arjunsv98:
Fair enough, you're right — I'd only checked the top-level tb parameters and missed configs reaching the demux through other DUTs. Swapped the tb for an elaboration-time assertion in `axi_demux_id_counters` (`2**CounterWidth > MaxTrans`). Verified it fires at time 0 with the original counter sizing and is silent with the fix.

src/axi_demux_simple.sv
```diff
@@ -65,7 +65,11 @@ module axi_demux_simple #(
   input  axi_resp_t   [NoMstPorts-1:0]  mst_resps_i
 );
 
-  localparam int unsigned IdCounterWidth = cc_pkg::idx_width(MaxTrans);
+  // The in-flight counters must be able to hold the value `MaxTrans` itself (0 to MaxTrans
+  // inclusive), so they are sized for `MaxTrans + 1` states.  Sizing them with
+  // `idx_width(MaxTrans)` lets them saturate at `MaxTrans - 1` when `MaxTrans` is a power of
+  // two, silently reducing the configured concurrency by one transaction (issue #249).
+  localparam int unsigned IdCounterWidth = cc_pkg::idx_width(MaxTrans + 1);
```
imchenwu:
```suggestion
  localparam int unsigned IdCounterWidth = cc_pkg::cnt_width(MaxTrans);
```

src/axi_demux_id_counters.sv
```diff
@@ -45,6 +49,15 @@ module axi_demux_id_counters #(
...
 
+// pragma translate_off
+  initial begin : validate_params
+    counter_holds_maxtrans : assert (2**CounterWidth > MaxTrans) else
+      $fatal(1, "CounterWidth (%0d bits, maximum value %0d) cannot represent MaxTrans (%0d): \
+                 the counters would report full at %0d in-flight transactions.",
+             CounterWidth, 2**CounterWidth - 1, MaxTrans, 2**CounterWidth - 1);
+  end
+// pragma translate_on
+
```
imchenwu: Maybe replace by runtime assertions, see the other comment.

src/axi_demux_id_counters.sv
imchenwu, line 152: Maybe use such runtime assertions to watch the number of in-flight transactions?
```suggestion
    cnt_full_means_limit: assert property(
      @(posedge clk_i) disable iff (~rst_ni)
      (cnt_full[i] |-> ({overflow, in_flight} >= MaxTrans))) else
        $fatal(1, "axi_demux_id_counters > Counter %0d reports full at %0d < MaxTrans = %0d.",
               i, {overflow, in_flight}, MaxTrans);
    cnt_no_push_beyond_limit: assert property(
      @(posedge clk_i) disable iff (~rst_ni)
      (push_en[i] |-> ({overflow, in_flight} < MaxTrans))) else
        $fatal(1, "axi_demux_id_counters > Counter %0d pushed at %0d >= MaxTrans = %0d.",
               i, {overflow, in_flight}, MaxTrans);
```

imchenwu (changes requested):

src/axi_demux_simple.sv
arjunsv98: TIL cnt_width exists — that's exactly the right helper here (and kind of poetic that the bug was idx_width being used where cnt_width belonged). Applied.

src/axi_demux_id_counters.sv
arjunsv98: Applied both as suggested and dropped the elaboration check.

imchenwu (approved):
Thx!

Issue #299, opened 2023-04-26
run_vsim.sh: Run simulations in parallel


colluca:
This is a feature request to speedup simulation and CI time. Simulations for different parameterizations of an IP could be run in parallel as independent processes.

colluca:
I think this would be a valuable feature (also to speed up the CI) and I would offer myself to implement this. The easiest would be to rewrite the `run_vsim.sh` script in Python. How do you feel about this? @micprog @thommythomaso

colluca:
Two comments to understand what's the plan for extending your Makefile, since you mention it's partial:
1) IMO it doesn't make sense to use Makefiles when all but the `*.log` targets are PHONY. And the only dependency of the `*.log` targets is basically the `Bender.yml` file not the actual HDL source files themselves, so most of the time you will end up force-running these rules as if they were PHONY. In this case, IMO sticking to the Makefile language (compared to another scripting language e.g. Python) is just more restrictive than the benefits it brings (since we are not using the useful features of Make most of the time).
2) You are still using the `run_vcs.sh` file with the loops parameterizing and simulating the MUT.

#### Regarding 1
I can extend your Makefile to add the source files of a specific TB and MUT as dependencies of the respective target. If I'm not mistaken, this should be easy and safe using Bender. This would really justify having the Makefile, as you simply do `make all` and he automatically figures out all that should be done.

#### Regarding 2
I wonder whether moving the loops into the Makefile could also make sense, each loop iteration would basically get its own simulation target. When you modify an HDL source file, all loop iterations would anyways run so there is no benefit from this point of view. From the parallelization point of view it would make sense. However the same could be achieved just modifying the `run_*.sh` scripts.

I thought about it and bringing this logic into the Makefile would probably be complex and criptic, with no reasonable advantage over the approach we are currently using. Here an example https://saveman71.com/2020/makefile-recursive-rules. In our case the cartesian product would be between the MUTs and their respective loop variables/bounds, with the additional complexity of not being a dense product.

So for this point I still propose to rewrite the `run_*.sh` scripts in Python and easily parallelize them there.

thommythomaso:
> Two comments to understand what's the plan for extending your Makefile, since you mention it's partial:
> 
[...]

Issue #432, opened 2026-07-02
Missing files in CI job dependency lists


colluca:
Hand-tracking module dependencies is error-prone, as demonstrated by the missing `axi_demux_id_counters` dependency in this list:
https://github.com/pulp-platform/axi/blob/e55ae2a7ee606ee3cfd4257f63982a971b704407/.gitlab-ci.yml#L243-L258

We should really consider going for an automated approach, perhaps leveraging Verilator, as e.g. done in Snitch:
https://github.com/pulp-platform/snitch_cluster/blob/2fa38482c2c822bfbedfdfd87abb3ed45521646e/make/common.mk#L104-L122

micprog:
Maybe `bender script flist -p axi -n --top <tb_module>` helps to get this list?

colluca:
Does that include also header files? If so, I guess it should do.

micprog:
I don't think header files propagate as they are (correctly) not in the file list. However, if using `flist-plus` you also get the required include dirs, and you can then check all files in those directories. It may not be as straightforward, but it does get to a result. Alternatively, use the bender pickle and hash the output, that should get exactly what is needed.

colluca:
I think `bender pickle` might make the most sense. I anyways suppose that what Github does behind the scenes is to calculate, if not a single, at least a per-file hash. And I don't expect there to be a significant performance difference between the two.

PR #442, opened 2026-08-05
scripts/run_vsim.sh: fix regression-script bugs and optimize vsim ci jobs for better speed and coverage.

### LINKED ISSUE #438: `run_vsim.sh`: duplicate `axi_xbar` case and other script issues
`scripts/run_vsim.sh` has two `axi_xbar)` entries in its `case` statement. Since bash only executes the first matching branch, the second one (added in 78f29996, sweeping ID width, `TbAxiIdUsed`, data width and pipeline) is dead code and has never run. These parameter sweeps are therefore missing from regression.

Minor issues in the same script:

- `axi_lite_regs` appends to the global `SEEDS` array, so in a full run every test executed after it is silently repeated with the extra seeds.
- `$VSIM ... | tee vsim.log 2>&1` redirects tee's stderr, not vsim's, so vsim stderr is not captured in the log.
- The unsupported-flag pattern `-*--*` does not match typical misspelled flags like `--foo`, which fall through and are treated as test names.
- The header comment mentions a `--random` flag, but the implemented flag is `--random-seed`.

Also as #299 mentioned, it would be great if we can find a way for vsim ci acceleration.

We might want to improve this vsim script in the next major release. @nikgiu

nikgiu:
scripts/run_vsim.sh: fix regression-script bugs and restore axi_xbar coverage

Addresses the issues reported in #438.

Fixed
Scope SEEDS to each test (local copy) so per-TB seed additions (e.g. axi_lite_regs) no longer leak into the subsequent tests of a full run.
Redirect vsim's stderr into the log by moving 2>&1 before tee (it previously captured tee's stderr instead).
Reject any unrecognized dash-prefixed flag (-*); the old -*--* pattern missed typos such as --foo, which were then misread as testbench names.
Update the stale --random reference in the header comment to --random-seed.

Changed
Merge the two duplicate axi_xbar case branches into one. The second branch (added in 78f2999) was shadowed by the first and never executed; the two parameter sweeps are complementary (exclusive-access / unique-id handling vs. ID-width usage, data width and pipelining), so both are now run.

Not included (follow-up)
This does not implement the parallel execution requested in #299: the axi_xbar merge increases wall-clock time, and any parallelization must be license-bounded (a fixed concurrency cap) to avoid exhausting the shared Questa license pool — left as a separate change.

### DIFF
--- a/scripts/run_vsim.sh
+++ b/scripts/run_vsim.sh
@@ -27,10 +27,50 @@ fi
 # stay regression-consistent.
 SEEDS=(0)
 
+# Maximum number of `vsim` invocations to run concurrently. Each running simulation occupies one
+# simulator license seat, so this also caps the license usage regardless of how many
+# parametrizations a test sweeps. Defaults to 1 (fully sequential, i.e. unchanged behaviour); set
+# e.g. `VSIM_JOBS=4` to parallelize. NOTE: values > 1 run multiple `vsim` processes in the same
+# directory sharing the compiled `work` library; each gets its own log/wlf, but validate on your
+# simulator before relying on it.
+: "${VSIM_JOBS:=2}"
+
+vsim_fail=0   # set to 1 as soon as any background simulation reports errors
+job_idx=0     # unique index per launched simulation, used for per-job artifact names
+
+# Block until fewer than VSIM_JOBS simulations are running.
+vsim_throttle() {
+    while (( $(jobs -rp | wc -l) >= VSIM_JOBS )); do
+        wait -n || vsim_fail=1
+    done
+}
+
+# Wait for all still-running simulations to finish.
+vsim_drain() {
+    local pid
+    for pid in $(jobs -rp); do
+        wait "$pid" || vsim_fail=1
+    done
+}
+
 call_vsim() {
-    for seed in ${SEEDS[@]}; do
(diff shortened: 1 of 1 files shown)


imchenwu (changes requested):
Hi Niccolo,

1. I feel it unsafe to use `jobs -rp` as it only looks at the running jobs, and if a job already faied at elaboration time, it won't be catched in `jobs -rp` and thus the failure log would be missing in the ci.
2. For tb_axi_xbar parameter sweep: maybe add the case when `IdUsed == IdWidth` for the sweep to have better coverage?
3. I think all the `+acc` can be dropped because it's a batch-only simulation.

imchenwu:
I did some modifications.

1. Dropped "+acc" and "+cover=".
2. Added missing files in CI job dependency lists as #432 reported.
3. Parallelized the `axi_xbar` ci job.

imchenwu:
I added some new optimizations here. The motivations:

1. Previously the pushes on the master branch won't trigger internal CI to run vsim jobs because the diff base is the newest commit on the master branch (which is itself in this case). Although most of the pushes on the master branch are verified by the CI checks in PRs, it is still safer to properly trigger CI again on the master branch.
2. On some branches under active development, all commits would trigger CI pipelines, and now new commits on those branches should cancel the previous pipelines because of this `auto_cancel`. The master branch is an exception for this `auto_cancel` feature.

